PR #7, opened 2023-03-31
Update README.md

rodolfocarobene:
There was a typo in the repository name and also the qick link was missing... 
If someone has other corrections we can add them here


AleCandido (approved):

AleCandido:
This kind of commits you could do even directly on `main` (if you have the rights to do it).

PR #9, opened 2023-04-05
ZCU111 server

rodolfocarobene:
This server builds on the one for single qubit, extending it.  

In particular this should add:
- [ ] Flux support (bias and pulses)
- [ ] Multiplexed readout
  - [ ] averaged results
  - [ ] single shots
- [ ] Solving everything also for sweepers

Before review:
- [ ] Tests for new functions
- [ ] Tests with qibocal routine via updated driver
- [ ] Documentation is updated


JavierSerranoGarcia:
@rodolfocarobene I got a couple of bugs when computing the Nyquist zones

```

def declare_nqz_zones(self, sequence: PulseSequence):
        ch_already_declared = []
        for pulse in self.sequence:
            if pulse.type is PulseType.DRIVE:
                ch = self.qubits[pulse.qubit].drive.ports[0][1]
            elif pulse.type is PulseType.READOUT:
                ch = self.qubits[pulse.qubit].readout.ports[0][1]
            elif pulse.type is PulseType.FLUX:
                ch = self.qubits[pulse.qubit].flux.ports[0][1]

            if ch not in ch_already_declared:
                ch_already_declared.append(ch)
                zone = 1 if pulse.frequency < self.max_sampling_rate / 2 else 2
                self.declare_gen(ch, nqz=zone)

```
and the other
    
```
 if adc_ch not in adc_ch_added:
                adc_ch_added.append(adc_ch)
                zone = 1 if pulse.frequency < self.max_sampling_rate / 2 else 2

                # TODO add parameters to QickProgramConfig
                freq = pulse.frequency - self.cfg["mixer_freq"] - self.cfg["LO_freq"]
```

in both cases you should substrate the LO and mixer frequencies (only for readout).

For drive pulses we should produce a warning in case the frequency is bigger than Fs

The same for this function

`   def declare_readout_freq(self):`

rodolfocarobene:
Thanks @JavierSerranoGarcia for spotting the error! It should be fixed now

src/qibosoq/qick_programs.py
```diff
@@ -16,7 +16,12 @@
 NS_TO_US = 1e-3
 
 
-class FluxProgram:
+class FluxProgram:  # TODO include and test
+    @staticmethod
+    # TODO use
+    def from_volt_to_gain(volt: float, p0: int = 329, p1: int = 25634):
+        return int(p0 + volt * p1)
```
AleCandido, line 23: Static methods are just a way to put something in a given namespace, since there is no reference to `self` or `cls`.

Consider if this does really belong to this namespace, or it can just be lifted top-level.

src/qibosoq/qick_programs.py
```diff
@@ -34,6 +39,8 @@ def set_bias(self, mode="sweetspot"):
                 i_wf = np.full(duration, value)
                 q_wf = np.zeros(len(i_wf))
                 self.add_pulse(ch, f"const_{value}_{idx}", i_wf, q_wf)
+
+                print(f"Setting bias of ch {ch} to gain {self.max_flux_gain}")
```
AleCandido, line 43: If possible avoid bare `print`s, and prefer [`logging`](https://docs.python.org/3/library/logging.html).

src/qibosoq/qick_programs.py
```diff
@@ -110,6 +118,7 @@ def __init__(
 
         # general settings
         self.max_gain = qpcfg.max_gain
+        self.max_flux_gain = 24000  # TODO not hardcode
```
AleCandido, line 121: I agree with the comment: in order not to forget (the `TODO` is already good in this sense), a better temporary solution is to set this variable as a global constant, and use the constant instead of the literal.

src/qibosoq/qick_programs.py
```diff
@@ -174,12 +184,15 @@ def declare_gen_mux_ro(self):
                 adc_ch_added.append(adc_ch)
                 # TODO add parameters to QickProgramConfig
                 freq = pulse.frequency - self.mixer_freq - self.LO_freq
-                zone = 1 if freq < self.max_sampling_rate / 2 else 2
+                max_sampling_rate = 1_536_000_000  # TODO this should be in config
```
AleCandido, line 187: See above (and considering that this appears twice, you can also immediately reuse the constant)

src/qibosoq/qick_programs.py
```diff
@@ -174,12 +184,15 @@ def declare_gen_mux_ro(self):
...
+                zone = 1 if freq < max_sampling_rate / 2 else 2
                 freq = freq * HZ_TO_MHZ
 
                 mux_gains.append(pulse.amplitude)
                 mux_freqs.append(freq)
 
+        print(f"declare gen {ro_ch} {zone} {self.mixer_freq} {mux_freqs} {mux_gains} {adc_ch_added[0]}")
```
AleCandido, line 194: Another `print`

src/qibosoq/qick_programs.py
```diff
@@ -290,19 +304,18 @@ def add_pulse_to_register(self, pulse: Pulse):
...
         gen_ch = self.qubits[pulse.qubit].readout.ports[0][1]
 
         us_length = pulse.duration * NS_TO_US
-        soc_length = self.soc.us2cycles(us_length)
+        soc_length = self.soc.us2cycles(us_length, gen_ch=gen_ch)
 
         if not isinstance(pulse.shape, Rectangular):
             raise Exception("Only Rectangular ro pulses are supported")
 
+        print(f"set pulse register {gen_ch} {soc_length} {mask}")
```
AleCandido, line 318: Another `print`

src/qibosoq/qick_programs.py
```diff
@@ -338,16 +350,18 @@ def body(self):
                         for ro_pulse in self.multi_ro_pulses[pulse.start]:
                             adcs.append(self.qubits[ro_pulse.qubit].feedback.ports[0][1])
 
+                        print(f"measure {ro_ch} {adcs} {time} {self.adc_trig_offset} {self.syncdelay}")
```
AleCandido, line 353: Another `print`

PR #10, opened 2023-04-11
Documentation

rodolfocarobene:
I never used sphinx or github actions so I'm not entirely sure that this is correct.
In particular I'm not sure if I added the correct github workflow: this is called "publish" but I think it directly load the documentation on the website and I'm not sure this is what I need to do.

In any case, this PR should end up adding the automatic sphinx documentation.

Checklist before merge:


Edoardo-Pedicillo (approved):
Thank @rodolfocarobene, LGTM

PR #11, opened 2023-04-13
Refactor

rodolfocarobene:
This PR tracks the refractoring of the first version of the server (also in preparation of the second version).
This should remove duplicated code and improve the readability, without introducing new features or fixing anything else.


rodolfocarobene:
This could already be reviewed, it does not change anything but the file and class structure.

In the end, a summary of the changes is:
* the big file that  was in main has been splitted in one dedicated to the server (so the TCPhandler and the main function) and in one dedicated only to qickprograms definition
* in qickprogram, now there is a GeneralQickProgram abstract class that is used to initialize both ExecutePulseSequence and ExecuteSingleSweep. This has removed a lot of duplication since the majority of the functions are the same.
* Some large functions (like `initialize`) have been splitted to improve the readability

The documentation was updated and the coverage is not decreasing (obviously, there are no tests in main for now, I will need to update the tests-PR)

AleCandido:
Two comments about the `if __name__ == "__main__":`:

1. it is better if you encapsulate the content in a `main()` function, since in this way it would also be possible to test
2. the actual execution makes your package an application, but it is rather weird if you then deploy it as a package: how would you run after installing with `pip install qibosoq`? (solution `python /path/to/env/lib/python3.x/site-packages/qibosoq/rfsoc_server.py`... not the best...)

Point 2. has an actual standard solution in Python: a `__main__.py` file inside your package. Same question: how would you run this after package installation? Solution: `python -m qibosoq`, much more convenient :)

I'm writing here to you since you're refactoring, but it is fine to keep this for a further PR. In case, just open a dedicated issue.

rodolfocarobene:
Thank you @AleCandido for the suggestions, I will open a new issue for the second point and solve the first here

AleCandido:
Btw, you have a consistent and interesting typo: ref**r**actor

src/qibosoq/__main__.py
```diff
@@ -0,0 +1,14 @@
...
+from qibosoq.rfsoc_server import MyTCPHandler
+
+HOST = "192.168.0.72"  # Server address
+PORT = 6000  # Port to listen on
+
+TCPServer.allow_reuse_address = True
+with TCPServer((HOST, PORT), MyTCPHandler) as server:
+    print("Server Listening")
+
+    server.serve_forever()
```
AleCandido: Wrap this in a function:
```suggestion
def serve(host, port):
    TCPServer.allow_reuse_address = True
    with TCPServer((host, port), MyTCPHandler) as server:
        print("Server Listening")

        server.serve_forever()
        
serve(HOST, PORT)
```

It will be easier to upgrade.

AleCandido: P.S.: the `serve()` function can also be located back in the `rfsoc_server` module :)

src/qibosoq/rfsoc_server.py
```diff
@@ -15,6 +15,11 @@
 
 from qibosoq.qick_programs import ExecutePulseSequence, ExecuteSingleSweep
 
+# initialize QickSoc object (firmware and clocks)
+global_soc = QickSoc()
```
AleCandido: Why do you need a global object?

In principle, you'd only need a single global object in your application, and that will be the handler.

Even more, if you use the `serve()` function, even that object will be local to `serve()`, and only the function call will be done in the global scope.

AleCandido: However, if you need it, independently of the handler, instantiate it inside `serve()`.

src/qibosoq/rfsoc_server.py
```diff
@@ -15,6 +15,11 @@
...
+# starts handler for system interruption (ex. Ctrl-C)
+signal.signal(signal.SIGINT, signal_handler)
```
AleCandido: Same of above: if needed, put it in `serve()`.

src/qibosoq/rfsoc_server.py
rodolfocarobene: Everytime QickSoc gets initialized it slowly re-loads the firmware, resetting clocks and changing the phase between DACs and ADCs, so it's important to not do it.

I understand that is better, in any case, to have it inside the serve function, but if I try I get an error:
```
File "/home/xilinx/qibosoq/src/qibosoq/rfsoc_server.py", line 45, in handle
    program = ExecutePulseSequence(global_soc, data["cfg"], data["sequence"], data["qubits"])
NameError: name 'global_soc' is not defined
```

At first I though to pass the QickSoc object as an attribute of MyTCPHandler/TCPServer, rewriting __init__ or adding a function, but I couldn't really do it and opted for the global

README.md
JavierSerranoGarcia, line 1: Maybe we could include how to run the mini webserver .... or we leave it for the ZCU111

JavierSerranoGarcia (approved):
Very nice code. Congrats

src/qibosoq/qick_programs.py
```diff
@@ -0,0 +1,364 @@
...
+from qibolab.sweeper import Parameter, Sweeper
+from qick import AveragerProgram, QickSoc, RAveragerProgram
+
+# conversion coefficients (in qibolab we use Hz and ns)
+HZ_TO_MHZ = 1e-6
+NS_TO_US = 1e-3
+
+
+class GeneralQickProgram(ABC):
+    """Abstract class for QickPrograms"""
```
stavros11, line 20: I am wondering if this class should inherit from some `qick` abstract class, since you are basically overriding/defining methods that qick requires and in some cases (`self.di_buf` or `self.us2cycles`) you are using qick methods in this class. 

I understand that you cannot inherit the `AveragerProgram` because you may need the `RAveragerProgram` instead, but maybe the base class of these. This may complicate the inheritance structure though and I am not sure how well the double inheritance below will work.

src/qibosoq/qick_programs.py
```diff
@@ -0,0 +1,364 @@
...
+        Args:
+            sequence (PulseSequence): sequence of pulses to consider
+        """
+
+        ch_already_declared = []
+        for pulse in self.sequence:
+            if pulse.type is PulseType.DRIVE:
+                ch = self.qubits[pulse.qubit].drive.ports[0][1]
+            elif pulse.type is PulseType.READOUT:
+                ch = self.qubits[pulse.qubit].readout.ports[0][1]
```
stavros11, line 66: If you have a better idea of how `ports` should look, feel free to propose. 
The current layout `tuple(controller number, port number)` was inspired by QM, so I am not sure if it is optimal for other devices. Indexing `[0][1]` all the time does not look good but we should also keep in mind that it will be more complicated when you have more qubits and controllers.

stavros11 (approved):
Well done for the refactoring!

Some comments below but mostly questions or suggestions, not really blocking. It would be good to consider whether `GeneralQickProgram` should inherit a qick class.

AleCandido (commented):
Looks like I had a pending review I forgot to submit...

If anything is outdated, please ignore it.

README.md
rodolfocarobene: I would maybe do this in a separate PR just for that. So that we do not mix things to much

src/qibosoq/qick_programs.py
```diff
@@ -0,0 +1,364 @@
...
+
+        adcs, adc_count = np.unique(adcs, return_counts=True)
+
+        for idx, adc_ch in enumerate(adcs):
+            count = adc_count[adc_ch]
+            if self.expts:  # self.expts is None if this is not a sweep
+                shape = (count, self.expts, self.reps)
+            else:
+                shape = (count, self.reps)
+            i_val = self.di_buf[idx].reshape(shape) / lengths[idx]
```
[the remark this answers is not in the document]
rodolfocarobene: I'll open an issue/pull request for pylint, indeed it would be useful

src/qibosoq/qick_programs.py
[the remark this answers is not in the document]
rodolfocarobene: I think I prefer the explicit version, I think it is more readable. But thanks for the suggestion, I did not know the getattr function

rodolfocarobene: I think that the current implementation is ok... in a very distant future when the driver will control multiple board at the same time I will use both attributes somewhere :-)

src/qibosoq/rfsoc_server.py
AleCandido: > The problem with this was that I do not have control on the MyTCPHandler object initialization. This is done inside the TCPServer class ([python/cpython@`7f70f1c`/Lib/socketserver.py#L361](https://github.com/python/cpython/blob/7f70f1c5cc964f64a89b3536b1bc96c3ecb6cb24/Lib/socketserver.py#L361)) and I didn't want to change that object at all (I could do it, but it seems a bit too much).

The suggestion above is really to extend the initialization of `TCPHandler` in `MyTCPHandler`, adding a simple step before, i.e. `soc` assignment.

With:
```py
class MyTCPHandler(BaseRequestHandler):
    def __init__(self, soc, *args, **kwargs):
        self.soc = soc
        super().__init__(*args, **kwargs)
```
you're simply saying:
- get all arguments that are passed
- the first one is special: assign to an attribute, and strip from the list
- pass down the rest of the arguments and execute the parent `__init__`, as if I did nothing and just received the same arguments but `soc`

This does not require patching CPython, it is just standard practice in subclassing. If you didn't want to modify the original class at all, you should not use subclassing, but composition:
```py
class MyTCPHandler:
    def __init__(self, soc, *args, **kwargs):
        self.soc = soc
        self.handler = BaseRequestHandler(*args, **kwargs)
```

But given that it is named `BaseRequestHandler` I really believe it is intentionally made for subclassing, and extending `__init__` has nothing special.

src/qibosoq/qick_programs.py
rodolfocarobene: Looking at pylint, this does not work as intended... The super.acquire function (called in the acquire function of GeneralQickProgram) should call the acquire either of AveragerProgram or of RAveragerProgram, but if the GeneralQickProgram inherits from QickProgram then the QickProgram.acquire function gets called...

If I don't add the QickProgram, however, it is true that pylint complains for self.di_buf, self.us2cycles and many more

AleCandido:
Have a look at comments on #15

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,45 @@
...
+        if: steps.cache.outputs.cache-hit != 'true'
+      - name: Install project
+        run: poetry install --no-interaction
+      - name: Install task runner
+        run: pip install poethepoet
+      - name: Lint
+        run: |
+          source $VENV
+          poe lint
+          poe lint-warnings
```
AleCandido, line 45: poethepoet (`poe`) is already making use of the correct environment, so you do not need activation.

Moreover, the `$VENV` variable is most likely unset, so you are sourcing nothing.

```suggestion
          poe lint
          poe lint-warnings
```

rodolfocarobene:
Thanks @AleCandido for the comments, everything should be corrected now. I had copied the main part of the code from  [qiboteam/workflows](https://github.com/qiboteam/workflows/blob/main/.github/workflows/rules-poetry.yml), maybe some of these problems can also be corrected there?

rodolfocarobene:
For the failing pylint and the inheritance chain, maybe its better if I explain a bit more... I don't think I'm able to solve this on my own, but it does not seem impossible.

Eventually I want to have this structure:
```python

class GeneralQickProgram(ABC, QickProgram):
  def __init__ (...):
    ...
    super().__init__(...)  # here it should call AveragerProgram/RAveragerProgram init

 def acquire (...):
    ...
    super().acquire(...)  # here it should call AveragerProgram/RAveragerProgram acquire

class ExecutePulseSequence(GeneralQickProgram, AveragerProgram):
  ...

class ExecuteSingleSweep(GeneralQickProgram, RAveragerProgram):
  ...
```
if I just add the QickProgram inheritance, however, the `super().__init__` calls the QickProgram init and same thing for acquire.


Also, pylint complains since the GeneralQickProgram does not know some of the attributes used just by the sweeper class and defined in its init... should I define them in GeneralQickProgram as None or is there a better way?

rodolfocarobene:
Actually is just a pylint problem... it sees the super init and acquire as the ones of QickProgram, but these have different numbers of parameters so it raises an error. 
I could just skip these lines in pylint (at least I could see the pylint output), but it's probably not the best idea ever

src/qibosoq/rfsoc_server.py
AleCandido: Ok, if you're not constructing it, then we can think about setting after initialization.

I.e.:
```py
class MyTCPHandler(BaseRequestHandler):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.soc = None
```
and then find a way to set `handler.soc = soc` afterward.

I need to have a look to the server, to see whether it exposes the handler, and how it gets access to the class.

rodolfocarobene:
I would like to merge this PR: this has already been reviewed and I think it's kinda ok. Also it's a bit easier to have it on main since every new PR had to start from here.

I will open a specific issue for the global soc problem

PR #15, opened 2023-04-27
pylint test

rodolfocarobene:
Adding pylint workflow

Checklist before merge:
- [ ] Tests are passing.


rodolfocarobene:
Pylint is complaining about some things, but I think the workflow itself is working..
However, I have not much experience with workflows so I'm not sure this is done correctly

rodolfocarobene:
Ok I see now that the tests are actually failing... Let me try to fix it

rodolfocarobene:
Ok now it should be working. The tests are failing because of the way I was trying to do abstraction but this is unrelated

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,45 @@
+
+
+name: Pylint
+
+on:
+  workflow_dispatch:
```
AleCandido, line 6: This is fine for development, but since you are running Pylint on every push, maybe you can eventually drop it.

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,45 @@
...
+  push:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
```
AleCandido, line 14: This is already the default, so you can drop it.

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8  # maybe better not hardcoded
```
AleCandido, line 20: You should not be so sensible to the Python version, so whatever is the default should be fine.

If instead you really want to stick to py3.8, because otherwise Pylint will enable some different lints (I don't think it would, but it's possible, I should check), then it is fine to hard-code, it doesn't make a big difference (in principle you could define a variable top-level, but it is even fine like this, do not overcomplicate things).

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Install and configure poetry
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+      - name: Disable modern installer
+        run: poetry config installer.modern-installation false
```
AleCandido, line 27: Do you really need to disable the modern installer? Most packages are working fine with it, and it speeds up installation.

Have a try without these lines (or even just try `poetry install` locally, after `poetry env remove python`, with a recent enough version of Poetry, i.e >= 1.4.1 and with this setting set to `true`, or just unset).

PR #17, opened 2023-04-28
Zcu111- flux

rodolfocarobene:
PR #9 was becoming difficult to handle, I was trying to implement both multiplex and  flux support in the same PR and this was probably a bad idea. Also, problem in multiplexing were stopping progress on flux.

This PR should just focus on writing a driver for 1 single flux-tunable qubit.

Checklist before merge:
- [ ] Tests are passing.


AleCandido:
I changed base to isolate which are the changes original in this PR, essentially to avoid reviewing twice #11.

If then you want to merge in a different order, that's perfectly fine by me.

src/qibosoq/__main__.py
```diff
@@ -27,9 +27,26 @@ def define_logger():
...
+    new_logger = logging.getLogger("qick_program")
+    new_logger.setLevel(logging.DEBUG)
+
+    filename = "/home/xilinx/logs/program.log"
+    formatter = logging.Formatter("%(levelname)s :: %(asctime)s\n%(message)s", "%Y-%m-%d %H:%M:%S")
+
+    handler = logging.handlers.RotatingFileHandler(filename, mode="w", backupCount=3)
+    handler.setFormatter(formatter)
+
+    new_logger.addHandler(handler)
```
AleCandido, line 42: Consider splitting this into a separate function, because it is just a function of `filename`, and it is almost the exact same of the one above.

src/qibosoq/__main__.py
```diff
@@ -27,9 +27,26 @@ def define_logger():
...
+    return new_logger
+
+
+logger = define_main_logger()
+program_logger = define_program_logger()
+
+HOST = "192.168.0.81"  # Server address
```
AleCandido, line 49: As said on the client side: this could be an [`ipadress`](https://docs.python.org/3/library/ipaddress.html)

src/qibosoq/abstracts.py
```diff
@@ -0,0 +1,65 @@
...
+
+from dataclasses import dataclass
+from enum import Enum, auto
+from typing import List, Union
+
+
+@dataclass
+class Config:
+    """General RFSoC Configuration"""
+
```
AleCandido, line 11: Usually, you don't put an empty line after the docstring (just conventions)
```suggestion
    """General RFSoC Configuration"""
```

src/qibosoq/abstracts.py
```diff
@@ -0,0 +1,65 @@
...
+    start = auto()
+
+    bias = auto()
+
+
+@dataclass
+class Sweeper:
+    """Sweeper object"""
+
+    expts: int = None  # single number of points
```
AleCandido, line 61: You can actually turn the comment into a docstring:
```suggestion
    expts: int = None
    """Single number of points."""
```

src/qibosoq/abstracts.py
AleCandido, line 1: More than `abstracts.py` (here there is no abstract class at all), I would call `data.py`, or something like.

src/qibosoq/__main__.py
```diff
@@ -27,9 +27,26 @@ def define_logger():
     return new_logger
 
 
-logger = define_logger()
+def define_program_logger():
+    """Define qick logger format and handler"""
 
-HOST = "192.168.0.72"  # Server address
+    new_logger = logging.getLogger("qick_program")
```
AleCandido, line 33: Move the name in a top-level constant, or ask it as an input of the function.

src/qibosoq/__main__.py
```diff
@@ -27,9 +27,26 @@ def define_logger():
...
+    new_logger.setLevel(logging.DEBUG)
+
+    filename = "/home/xilinx/logs/program.log"
```
AleCandido, line 36: Same of above.

src/qibosoq/rfsoc_server.py
```diff
@@ -1,20 +1,28 @@
...
+import traceback
 from socketserver import BaseRequestHandler, TCPServer
 
 from qick import QickSoc
 
-from qibosoq.qick_programs import ExecutePulseSequence, ExecuteSingleSweep
+from qibosoq.abstracts import Config, Pulse, Qubit, Sweeper
+from qibosoq.qick_programs import ExecutePulseSequence, ExecuteSweeps
 
 logger = logging.getLogger("__name__")
```
AleCandido, line 20:
```suggestion
logger = logging.getLogger(__name__)
```

[`__name__`](https://docs.python.org/3/reference/import.html#__name__) is a special variable.

src/qibosoq/rfsoc_server.py
```diff
@@ -42,12 +50,26 @@ def execute_program(self, data: dict) -> dict:
...
-            program = ExecutePulseSequence(global_soc, data["cfg"], data["sequence"], data["qubits"])
-        elif data["operation_code"] == "execute_single_sweep":
-            program = ExecuteSingleSweep(global_soc, data["cfg"], data["sequence"], data["qubits"], data["sweeper"])
+            program = ExecutePulseSequence(
+                global_soc,
+                Config(**data["cfg"]),
+                [Pulse(**pulse) for pulse in data["sequence"]],
+                [Qubit(**qubit) for qubit in data["qubits"]],
+            )
+        elif data["operation_code"] == "execute_sweeps":
```
AleCandido, line 59: Operation codes could be an `enum`

src/qibosoq/rfsoc_server.py
```diff
@@ -42,12 +50,26 @@ def execute_program(self, data: dict) -> dict:
...
+            program = ExecuteSweeps(
+                global_soc,
+                Config(**data["cfg"]),
+                [Pulse(**pulse) for pulse in data["sequence"]],
+                [Qubit(**qubit) for qubit in data["qubits"]],
```
AleCandido, line 64: These lines look exactly like those above :)

src/qibosoq/rfsoc_server.py
```diff
@@ -66,20 +88,23 @@ def handle(self):
...
         # set the server in non-blocking mode
         self.server.socket.setblocking(False)
 
         try:
+            logger.debug("Receiving data")
             data = self.receive_command()
+            logger.debug("Executing program")
             results = self.execute_program(data)
-        except Exception as exception:
+        except Exception as exception:  # pylint: disable=bare-except, broad-exception-caught
```
AleCandido, line 100: Do you really need such a broad exception?

src/qibosoq/qick_programs.py
```diff
@@ -174,42 +152,78 @@ def add_pulse_to_register(self, pulse: Pulse):
...
+        """Check if two pulses are equal, does not check the start time"""
+        if pulse_a is None or pulse_b is None:
+            return False
+        return (
+            pulse_a.frequency == pulse_b.frequency
+            and pulse_a.amplitude == pulse_b.amplitude
+            and pulse_a.relative_phase == pulse_b.relative_phase
+            and pulse_a.duration == pulse_b.duration
+            and pulse_a.type == pulse_b.type
+        )
```
stavros11, line 224: This function can also be defined as the `Pulse.__eq__` instead of putting it in the `BaseProgram`. Note that as it is now it is not using the program (`self`) anyway. 

I also cannot find where it is being used. If it's not maybe it could even be dropped, otherwise, you can replace it by `pulse_a == pulse_b` after defining it as `__eq__`.

src/qibosoq/qick_programs.py
```diff
@@ -228,11 +242,11 @@ def acquire(
             debug (bool): if true prints the program actually executed
             average (bool): if true return averaged res, otherwise single shots
         """
-
+        if self.readouts_per_experiment is not None:
+            readouts_per_experiment = self.readouts_per_experiment
```
stavros11, line 246: Here you are overriding the `readouts_per_experiment` the user gives as an argument when calling the function. Is this really what you want?

src/qibosoq/qick_programs.py
```diff
@@ -228,11 +242,11 @@ def acquire(
...
         # if there are no readouts, temporaray set 1 so that qick can execute properly
         reads_per_rep = 1 if readouts_per_experiment == 0 else readouts_per_experiment
```
stavros11, line 248: Here you can directly put
```py
if readouts_per_experiment == 0:
    return [], []
```
to avoid calling the qick `acquire` when `readouts_per_experiment == 0`.

src/qibosoq/qick_programs.py
rodolfocarobene: Yes, it ovverides the number of readout only in the case of multiplexed readout because 3 acquisition at the same time are considered by qick a single acquisition, but in qibolab (and in general considering pulse sequences) are multiple pulses.

src/qibosoq/qick_programs.py
rodolfocarobene: Actually I want to execute the program even if there are no readouts, to eventually just drive the qubit... Not sure if it is something useful

src/qibosoq/rfsoc_server.py
rodolfocarobene: I don't know everything of socket and qick libraries, but I want that whatever error get raised, by whatever library:
* the client gets to see it
* the server does not stop
Not sure if there is a better way

src/qibosoq/qick_programs.py
```diff
@@ -559,9 +542,9 @@ def initialize(self):
...
 
 SWEEPERS_TYPE = {
-    Parameter.frequency: "freq",
-    Parameter.amplitude: "gain",
-    Parameter.bias: "gain",
-    Parameter.relative_phase: "phase",
-    Parameter.start: "t",
+    Parameter.FREQUENCY: "freq",
+    Parameter.AMPLITUDE: "gain",
+    Parameter.BIAS: "GAin",
```
AleCandido, line 547: Maybe the case is incorrect, isn't it?
```suggestion
    Parameter.BIAS: "Gain",
```

src/qibosoq/rfsoc_server.py
rodolfocarobene: But what can i do? I would like to do something similar to 
```python
if sequence:
    program = ExecutePulseSequence
elif sweepers:
    program = ExecuteSweeps

program = program ( soc, Config(**data["cfg"], ...)
```
but in the case of sweepers, the program takes one variable more and I do not know what to do

src/qibosoq/rfsoc_server.py
```diff
@@ -49,14 +49,14 @@ def execute_program(self, data: dict) -> dict:
         Returns:
             (dict): dictionary with two keys (i, q) to lists of values
         """
-        if data["operation_code"] == "execute_pulse_sequence":
+        if OperationCode(data["operation_code"]) is OperationCode.EXECUTE_PULSE_SEQUENCE:
```
AleCandido, line 52:
```suggestion
        opcode = OperationCode(data["operation_code"])
        if opcode is OperationCode.EXECUTE_PULSE_SEQUENCE:
```

Just to save the repeated operation.

src/qibosoq/__main__.py
rodolfocarobene: But I use the name of the logger in another file (`rfsoc_server.py`). 
Should I define all the constants in `__init__.py` and import them around?

AleCandido: > Should I define all the constants in `__init__.py` and import them around?

You got the issue. For the time being, just lift to a separate file the constants that you use in multiple places.

I'd suggest avoiding `__init__.py`, because you should have a one-directional flow to avoid circular dependencies.
So, my advice is to `import` only from siblings or descendants (and transitively from descendants of siblings, of course), but not from parents or other files that are not in the same subtree (sometimes is better to give up on this last part, and allow imports from other subtrees, but with a clear hierarchy in mind between them).

So:
```
A/
 b/
  c # can import only from `d`
  d # can import from `c`, if `c` is not importing from it
 e/
  f # can import from `g` and `h/i`
  g # (if really needed, can also import `../b/c`, but then better to
       make sure that no one from inside `b` is importing anything from
       `e`)
  h/
   i # <- can not import from anyone else
```

And about `__init__.py`, you always have to consider it a level above the other modules in the same folder, because it represents the folder itself.

So, I would not put in the global `__init__.py`, in order to keep that just for the global exports.

rodolfocarobene: Actually this is a problem also for the main logger, since `__name__` is not a constant between the files... Didn't see it before because of `"__name__"`

src/qibosoq/rfsoc_server.py
AleCandido: There are many solutions, the complicate part is just to find the simplest one.

A possible one can be:
```py
if sequence:
    progcls = ExecutePulseSequence
    args = ()
elif sweepers:
    progcls = ExecuteSweeps
    args = ([Sweeper(**sweeper) for sweeper in data["sweepers"]],)

program = progcls(soc, Config(**data["cfg"]), ..., *args)
```
but it is not necessarily the best one.

AleCandido:
@rodolfocarobene, a frequent approach for logging is setting the configurations globally, and actually log in each file on `__name__`.
I would suggest doing this for the main logger, you can set the file where to log globally (they will all use the same). You can set the global configurations even `__main__.py`, as you were doing, since it is the entry point: once set there, they will be available for all the loggers.

However, if you want to have a separate logger, you can consider two alternatives:
- if used in a single file, you could define there
- otherwise, you could use a `log.py` file, and import from this file

The `configure_logger()` function could be defined in `log.py` in any case, since it will be reused by both the loggers anyhow.

rodolfocarobene:
@AleCandido I added a log.py file with the two log-related functions. 
I didn't yet defined the global logging and use it with `__name__`... Is there an advantage of using it or it's a convention? I suppose that the main advantage is to be able to see which file is logging what, but in this project it doesn't seem particularly useful.

doc/source/getting-started/usage.rst
```diff
@@ -1,30 +1,31 @@
 Usage instructions
 ==================
 
-After the installation, in ``<qibosoq-folder>/src/qibosoq`` you will find the ``__main__.py`` file.
-Here, the IP and port of the server are hardcoded and modifiable (if you installed qibosoq in developer mode).
+After the installation, in ``<qibosoq-folder>/src/qibosoq`` you will find the ``configuration.py`` file.
+Here, various qibosoq parameter are hardcoded and modifiable (if you installed qibosoq in developer mode).
```
AleCandido, line 5: We can also treat them as default parameters, and modify with environment variables (such that you don't have to modify the source).

src/qibosoq/__main__.py
rodolfocarobene: Sorry I didn't comment on this suggestion, seems a bit of an overkill to me... 
In the end if someone tries to set something that is not valid it will just fail

doc/source/getting-started/usage.rst
AleCandido:
```sh
❯ export CIAO=3
❯ sudo echo $CIAO
3
❯ python3 -c "import os; print(os.environ['CIAO'])"
3
❯ sudo python3 -c "import os; print(os.environ['CIAO'])"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "<frozen os>", line 679, in __getitem__
KeyError: 'CIAO'
❯ sudo -E python3 -c "import os; print(os.environ['CIAO'])"
3
```

src/qibosoq/rfsoc_server.py
```diff
@@ -42,26 +46,37 @@ def execute_program(self, data: dict) -> dict:
...
-            program = ExecutePulseSequence(global_soc, data["cfg"], data["sequence"], data["qubits"])
-        elif data["operation_code"] == "execute_single_sweep":
-            program = ExecuteSingleSweep(global_soc, data["cfg"], data["sequence"], data["qubits"], data["sweeper"])
+        opcode = OperationCode(data["operation_code"])
+        args = ()
+        if opcode is OperationCode.EXECUTE_PULSE_SEQUENCE:
+            programcls = ExecutePulseSequence
+        elif opcode is OperationCode.EXECUTE_SWEEPS:
+            programcls = ExecuteSweeps
+            args = (Sweeper(**sweeper) for sweeper in data["sweepers"])
```
AleCandido, line 55: Pay attention to this: this is not a tuple, but a generator (essentially an inline syntax to define a function that yields).

I like it a lot, but might become counter-intuitive.

In particular, being essentially a function, it is a closure: it captures variables from the environment it is defined in, but it uses it only when it is iterated.

```py
>>> l = list(range(5))
>>> f = (i for i in l)
>>> l.append(10)
>>> list(f)
[0, 1, 2, 3, 4, 10]
```

If you want a tuple instead, just iterate it immediately:
```suggestion
            args = tuple(Sweeper(**sweeper) for sweeper in data["sweepers"])
```

doc/source/getting-started/usage.rst
rodolfocarobene: nice, thanks!

rodolfocarobene:
Hi @AleCandido, thanks for all the reviews (here and in qibolab). Is something still missing?
I wanted to wait for approval before testing it also with the rfsoc4x2 :-)

src/qibosoq/components.py
```diff
@@ -0,0 +1,107 @@
...
+    """Can be 'rectangular', 'gaussian', 'drag'"""
+    rel_sigma: float = None
+    """Sigma for gaussians and drags, fraction of duration"""
+    beta: float = None
+    """Beta for drag pulses"""
+
+    def __eq__(self, pulse):
+        """Check if two pulses are equal, ignoring start time"""
+        if pulse is None:
+            return False
```
AleCandido, line 72: This is already the default in the automatic implementation (I know you needed here, but see the next comment).

src/qibosoq/components.py
```diff
@@ -0,0 +1,107 @@
...
+        return (
+            self.frequency == pulse.frequency
+            and self.amplitude == pulse.amplitude
+            and self.relative_phase == pulse.relative_phase
+            and self.duration == pulse.duration
+            and self.type == pulse.type
+            and self.dac == pulse.dac
+            and self.shape == pulse.shape
```
AleCandido, line 80: Instead of enumerating the fields you want to compare, you can rely on the automatic implementation, just telling it which fields you want to skip:
```py
beta: float = field(default=None, compare=False)
```
https://docs.python.org/3/library/dataclasses.html#dataclasses.field

src/qibosoq/configuration.py
```diff
@@ -0,0 +1,39 @@
+"""Qibosoq configuration file"""
+import os
+
+HOST = "192.168.0.81"
+"""Server address"""
+if "QIBOSOQ_HOST" in os.environ:
+    HOST = os.environ["QIBOSOQ_HOST"]
```
AleCandido, line 7:
```suggestion
def from_env(name, default=None):
    return os.getenv(f"QIBOSOQ_{name}", default)

HOST = from_env("HOST", "192.168.0.81")
"""Server address"""
```

AleCandido: In principle, we could remove even more duplication with a "dirty trick", that limited to this context would not be dirty:
```suggestion
_g = globals()
def from_env(name, default=None):
    _g[name] = os.getenv(f"QIBOSOQ_{name}", default)

from_env("HOST", "192.168.0.81")
```

But I would not recommend: generating dynamically we would lose Sphinx docs (there is no variable to assign to, since it is dynamically created) and it would be harder to understand.
I just wanted to show you that is possible (and someone else is doing similar but [better things](https://docs.pydantic.dev/latest/usage/settings/), but the first suggestion is good enough, we do not need the extra dependency, nor the extra complications),

---

You can test yourself with:
```sh
python -c "g = globals(); g['ciaocome'] = 42; print(42)"
```

src/qibosoq/rfsoc_server.py
```diff
@@ -1,21 +1,25 @@
...
 import os
-import pickle
 import socket
+import traceback
 from socketserver import BaseRequestHandler, TCPServer
 
 from qick import QickSoc
 
-from qibosoq.qick_programs import ExecutePulseSequence, ExecuteSingleSweep
+import qibosoq.configuration as qibosoq_cfg
```
AleCandido, line 17:
```suggestion
import qibosoq.configuration as cfg
```

You're in `qibosoq`, `cfg` should be good enough ;)

AleCandido (commented):
I will have a last look at the `qick_programs.py` module later on. However, in general it is good enough, so I'm just postponing the approval for a few hours.

Here just some marginal suggestions (cosmetics, but possibly improving maintanability, as usual).

src/qibosoq/components.py
rodolfocarobene: Thanks for the suggestion!

src/qibosoq/configuration.py
rodolfocarobene: Still nice to know, thanks! (I implemented the first version, mainly for having better documentation)

src/qibosoq/qick_programs.py
```diff
@@ -33,62 +29,62 @@ def __init__(self, soc: QickSoc, qpcfg: QickProgramConfig, sequence: PulseSequen
...
                 self.declare_gen(gen_ch, nqz=zone)
 
     def declare_readout_freq(self):
         """Declare ADCs downconversion frequencies"""
-
         adc_ch_already_declared = []
-        for readout_pulse in self.sequence.ro_pulses:
-            adc_ch = self.qubits[readout_pulse.qubit].feedback.ports[0][1]
-            ro_ch = self.qubits[readout_pulse.qubit].readout.ports[0][1]
+        for readout_pulse in [pulse for pulse in self.sequence if pulse.type == "readout"]:
```
AleCandido, line 80:
```suggestion
        for readout_pulse in (pulse for pulse in self.sequence if pulse.type == "readout"):
```

No practical difference, but just to let you know: this is the optimal place for a generator, because in this way the iteration will only happen once (during the `for` loop), just filtering the iterator.
Instead, if you use list comprehension, it would be iterated twice: first to construct the list (that is only used once, and then thrashed), and then for the `for` loop.

Since you're not modifying the elements, in this case another option would be:
```suggestion
        for readout_pulse in filter(lambda p: p.type == "readout", self.sequence):
```

In this case they are both fine by me. In general, comprehension is more powerful, because it's the combination of `map()` and `filter()`, with an elegant syntax. But since you only need `filter()`, here they are both fine.

AleCandido (approved):
Ok, this PR is already providing good enough results, I'd merge it.

Later on, we should review `qick_programs.py`, which is growing, and it's at the core of `qibosoq`. But not now :)

PR #23, opened 2023-05-09
Avoid unnecessary set_pulse_register

rodolfocarobene:
This should help in the flipping problem, where we probably reached the maximum number of instruction per program, and overall performances should also improve.

I did not have the opportunity to test it.

With this PR qibosoq will:
- [x] Not call `set_pulse_register` when not necessary
- [x] Qick supports adding multiple_pulses **before** firing them, this should also improve performances
- [x] Adding new log file with last executed qick_program in asm (this would make debug and communications with qick easier)

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/src/qibosoq/qick_programs.py
+++ b/src/qibosoq/qick_programs.py
@@ -51,6 +51,7 @@ class GeneralQickProgram(ABC, QickProgram):
         self.wait_initialize = self.us2cycles(2.0)
 
         self.pulses_registered = False
+        self.registered_waveform = []
 
         # pylint: disable-next=too-many-function-args
         super().__init__(soc, asdict(qpcfg))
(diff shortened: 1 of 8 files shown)


rodolfocarobene:
This has been deployed now on the board, I'll wait for review and some testing before merge

src/qibosoq/qick_programs.py
```diff
@@ -210,6 +219,19 @@ def body(self):
...
+        if pulse_a is None or pulse_b is None:
+            return False
+        return (
+            pulse_a.frequency == pulse_b.frequency
+            and pulse_a.amplitude == pulse_b.amplitude
+            and pulse_a.relative_phase == pulse_b.relative_phase
+            and pulse_a.duration == pulse_b.duration
+            and pulse_a.type == pulse_b.type
+        )
+        # and pulse_a.shape == pulse_b.shape
```
rodolfocarobene, line 233: hi @stavros11, in this function I wanted to compare 2 pulses (but not the start time). 
Everything worked fine, expect the comparison between shapes that, for identical rectangular pulses was giving
`pulse_a.shape == pulse_b.shape` equal to `False`.
Do you know what could have cause it?

rodolfocarobene:
@JavierSerranoGarcia could you have a look at this? If it's ok we can merge and reboot the server today/tomorrow

JavierSerranoGarcia (approved):
Looks everything OK

PR #31, opened 2023-06-07
Modify pre-commits

rodolfocarobene:
So, I added pycln and also a pre-commit to remove print statements (that removes the print tests)

Checklist before merge:
- [ ] Tests are passing.


.pre-commit-config.yaml
```diff
@@ -23,3 +23,17 @@ repos:
     rev: v3.4.0
     hooks:
       - id: pyupgrade
+  - repo: https://github.com/dhruvmanila/remove-print-statements
```
AleCandido, line 26: Pay attention, this doesn't seem super-maintained (little activity, unknown adoption...)

It might be fine, but we should drop it as soon as it will cause issues

rodolfocarobene: If it will cause issues I'll maybe try to link the pre-commit to the test_print.py... I think it's better if that check is done before committing

AleCandido: I agree that it is a better solution, also because the other one consisted I'm copy-pasting the same test across different repos.

In case something goes wrong, we can always fork that repo and fix it by ourselves (that is still easier than making the hook ourselves).

I just wanted to warn you about this option, since it might not work forever out of the box. But I agree that in the current situation it's a suitable solution.

AleCandido (approved):

PR #32, opened 2023-06-09
Enable raw acquisition

rodolfocarobene:
Checklist before review:
- [x] Raw acquisition working (test with time-of-flight routine in qibocal)
- [ ] Add tests
- [x] Check exceptions (what happens with more than one readout?)

Checklist before merge:
- [x] Reviewers confirm new code works as expected.
- [ ] Tests are passing.

### DIFF
--- a/src/qibosoq/rfsoc_server.py
+++ b/src/qibosoq/rfsoc_server.py
@@ -71,12 +71,14 @@ class ConnectionHandler(BaseRequestHandler):
         qick_logger.info(program.asm())
 
         if opcode is OperationCode.EXECUTE_PULSE_SEQUENCE_RAW:
-            toti, totq = program.acquire_decimated(
+            results = program.acquire_decimated(
                 self.server.qick_soc,
                 load_pulses=True,
                 progress=False,
                 debug=False,
             )
+            toti = [[results[0][0].tolist()]]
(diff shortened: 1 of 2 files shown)


src/qibosoq/rfsoc_server.py
```diff
@@ -66,16 +70,28 @@ def execute_program(self, data: dict) -> dict:
...
-
-        return {"i": toti.tolist(), "q": totq.tolist()}
+        if opcode is OperationCode.EXECUTE_PULSE_SEQUENCE_RAW:
+            results = program.acquire_decimated(
+                self.server.qick_soc,
+                load_pulses=True,
+                progress=False,
+                debug=False,
+            )
+            toti = [[results[0][0].tolist()]]
```
JavierSerranoGarcia, line 80: The raw data can be quite large. We have to check the TCP transmissions to give no problems with large amounts of data.

JavierSerranoGarcia (approved):

src/qibosoq/rfsoc_server.py
rodolfocarobene: For what I saw, the server automatically splits large packets in smaller ones so this *shouldn't* be a problem

PR #37, opened 2023-06-15
Fix config object

rodolfocarobene:
I added soft_avgs for raw acquisition, but if it's defined as None then will throw errors. There is no need of having the parameter there in any case, I will simply add it to the dictionary if needed.

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/src/qibosoq/qick_programs.py
+++ b/src/qibosoq/qick_programs.py
@@ -460,10 +460,7 @@ class ExecuteSweeps(FluxProgram, NDAveragerProgram):
     ):
         """Init function, sets sweepers parameters before calling super.__init__."""
         # sweepers are handled by qick in the opposite order
-        if isinstance(sweepers, Sweeper):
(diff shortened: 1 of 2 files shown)


src/qibosoq/qick_programs.py
```diff
@@ -460,7 +460,10 @@ def __init__(
     ):
         """Init function, sets sweepers parameters before calling super.__init__."""
         # sweepers are handled by qick in the opposite order
-        self.sweepers = list(sweepers)[::-1]
+        if isinstance(sweepers, Sweeper):
+            self.sweepers = [sweepers]
+        else:
+            self.sweepers = list(sweepers)[::-1]
```
Jacfomg, line 466: self.sweepers = [sweepers] if isinstance(sweepers, Sweeper) else list(sweepers)[::-1]

Jacfomg (approved):

PR #39, opened 2023-06-18
Check memory of tproc is not filled, raise informative error

rodolfocarobene:
I tried to run standard RB (with a non-calibrated qubit, just to see if the memory problem was solved) and it's not.
As I explained some time ago, there are two memories that can cause problems if filled: the DAC memory (that stores waveforms) and the tproc memory (that store the program). 
I solved the DAC memory problem for a standard program that doesn't use too many different waveforms, but the tproc still has the same limits. 
Note that the problem affects the two boards in very different ways: the RFSoC4x2 has 1024 memory limit (around 70 pulses), while the ZCU111 has a 8192 memory (around 560 pulses).

I have an idea to increase the maximum number of pulses, but in qibosoq it would probably appear as a weird patch. I am thinking about directly opening an issue and PR in qick (but maybe in the future).

In the moment, I added a check for the tproc memory so that eventually a much more informative error gets raised.

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/src/qibosoq/rfsoc_server.py
+++ b/src/qibosoq/rfsoc_server.py
@@ -71,7 +71,7 @@ class ConnectionHandler(BaseRequestHandler):
         qick_logger.handlers[0].doRollover()
         qick_logger.info(asm_prog)
 
-        num_instructions = asm_prog.count("\n") - 2
+        num_instructions = len(program.prog_list)
         max_mem = self.server.qick_soc["tprocs"][0]["pmem_size"]
         if num_instructions > max_mem:
             raise MemoryError(
--- a/src/qibosoq/rfsoc_server.py
+++ b/src/qibosoq/rfsoc_server.py
@@ -67,8 +67,16 @@ class ConnectionHandler(BaseRequestHandler):
             *args,
         )
 
+        asm_prog = program.asm()
         qick_logger.handlers[0].doRollover()
-        qick_logger.info(program.asm())
+        qick_logger.info(asm_prog)
+
+        num_instructions = asm_prog.count("\n") - 2
+        max_mem = self.server.qick_soc["tprocs"][0]["pmem_size"]
+        if num_instructions > max_mem:
+            raise MemoryError(
+                f"The tproc has a max memory size of {max_mem}, but the program had {num_instructions} instructions"
+            )
 
         if opcode is OperationCode.EXECUTE_PULSE_SEQUENCE_RAW:
             results = program.acquire_decimated(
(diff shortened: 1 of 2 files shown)


src/qibosoq/rfsoc_server.py
```diff
@@ -67,8 +67,16 @@ def execute_program(self, data: dict) -> dict:
             *args,
         )
 
+        asm_prog = program.asm()
         qick_logger.handlers[0].doRollover()
-        qick_logger.info(program.asm())
+        qick_logger.info(asm_prog)
+
+        num_instructions = asm_prog.count("\n") - 2
```
AleCandido, line 74: This should be equivalent to:
```suggestion
        num_instructions = len(program.prog_list) - 2
```
https://github.com/openquantumhardware/qick/blob/d393103ff324265bd33d4ba3fcb25a4ef3534101/qick_lib/qick/qick_asm.py#L2382

The advantage is that you do not have to iterate a whole string to obtain this information. And in principle you don't have to iterate at all, since `list`s knows their length.

But test it before using.

AleCandido (approved):
Apart from a tiny suggestion, it looks good to me :)

AleCandido:
> I have an idea to increase the maximum number of pulses, but in qibosoq it would probably appear as a weird patch. I am thinking about directly opening an issue and PR in qick (but maybe in the future).

Working directly on Qick will open a lot of options (including a bit of refactoring of their code...), but maybe this one in particular might not be a priority.

My current suggestions would be to start maintaining a list of potential improvements. Then, at some point we can decide to commit to that, and we will have a broader look, and make some planning to establish priorities (after that you can do whatever you feel more relevant by yourself... as you can do even now, but diving into someone else's code is expensive, that's why I'd recommend a bit of planning ahead of time).

src/qibosoq/rfsoc_server.py
rodolfocarobene: Nice, I was looking for something similar in qick but didn't find it... thanks!

rodolfocarobene:
> Working directly on Qick will open a lot of options (including a bit of refactoring of their code...), but maybe this one in particular might not be a priority.
I think that they (or he, since there is basically a single contributor) are trying to do some refactoring

> My current suggestions would be to start maintaining a list of potential improvements. Then, at some point we can decide to commit to that, and we will have a broader look, and make some planning to establish priorities (after that you can do whatever you feel more relevant by yourself... as you can do even now, but diving into someone else's code is expensive, that's why I'd recommend a bit of planning ahead of time).

Yes I agree, because of that I will not do anything for now :-)
But sooner or later, maybe when qibosoq will have all the supported features implemented, I think we should start working also on that. To speed up the development

PR #28, opened 2023-06-02
Fix Drag pulse

rodolfocarobene:
Checklist before review:
- [x] Drag pulse with beta=0 produces same AllXY than gaussian
- [x] AllXY with beta caracterization gives sensible results

Checklist before merge:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Documentation is updated.

PR #41, opened 2023-06-19
Function and test for preparing the sweepers list

### LINKED ISSUE #40: Unexpected error that I can't reproduce(?)
Hello @stavros11 @AleCandido, sorry to bother you but I need your expertise :-)
Is this line wrong for some reason?

https://github.com/qiboteam/qibosoq/blob/2e46ec70af79d04dcd747b3f8ead0ed9c3989a74/src/qibosoq/qick_programs.py#L463

I was getting an error from this line saying that `Sweeper object is not iterable` so I stopped the server, added some log.info before (printing sweepers and type(sweepers) I think) and the error disappeared. Than I removed the log.info and now there is no error...
Is this line of code "unsafe" for some reason or have you any idea of why/how something like this could have happened?

rodolfocarobene:
Fixes #40 by adding a separate function for the conversion/inversion and tests.

Checklist before merge:
- [ ] Tests are passing.


rodolfocarobene:
@AleCandido this is ready for review. I also added here `sweeper.parameter .>sweeper.parameters` so it's a bit more explicit (and *right*). When this is merged to main I will also update in qibolab.

After this, I'll focus a bit on the characterization so that I can test the fix for the drag pulse and, if I am able to solve it, I think we will be ready for the release... Not sure if you think something else still must be improved

AleCandido:
Thanks @rodolfocarobene, I'll review this today.

In general, out of `qick_programs.py`, `qibosoq` is perfect and simple. The only minor improvement could be splitting `ConnectionHandler.execute_program` out of the server, to separate the server itself from the task, putting it in a different module as a function (the only thing it needs of `self` is `self.server.qick_soc`, and it could be passed as an argument). Actually, in multiple functions. But that's a minor improvement.

A quick improvement could be to split `qick_programs.py` in a subpackage:
```
programs/
  __init__.py
  base.py
  flux.py
  pulse_sequence.py
  sweepers.py
```

The easy way suggested above consists in splitting each class to its own module.

Then we could try to improve the classes themselves, e.g. reducing the level of nesting and defining more atomic methods/functions. But this is something that will require continuous improvements, so there is no hurry.

Definitely #28 and #29 are waiting since quite some time :)

AleCandido (commented):
It seems like I forgot to submit a review... as usual...

src/qibosoq/components.py
```diff
@@ -80,14 +81,33 @@ class Parameter(IntEnum):
...
     DURATION = auto()
 
+    @overload
+    @classmethod
+    def variants(cls, parameters: str) -> "Parameter":
+        """Convert a string to a Parameter."""
+
+    @overload
+    @classmethod
+    def variants(cls, parameters: Iterable[str]) -> Iterable["Parameter"]:
```
AleCandido, line 91: Just saying `Iterable` it's a restricted version of the `Union`: you know that given an iterable, you'll obtain back another one, but it's not telling you that is an iterable *of the same kind*... but there is no better option 😞 

Unfortunately, when I wrote it I used a feature that is not a Python feature, even though there are actually people advocating for it
https://github.com/python/typing/issues/548

src/qibosoq/components.py
```diff
@@ -80,14 +81,33 @@ class Parameter(IntEnum):
...
+        """Convert a iterable of str to an iterable of Parameters."""
+
+    @classmethod
+    def variants(cls, parameters):
+        """Convert from strings to Parameters."""
+        if isinstance(parameters, str):
+            return cls[parameters.upper()]
+        if isinstance(parameters, Iterable):
+            return type(parameters)(cls[par.upper()] for par in parameters)
+        raise NotImplementedError("Conversion not supported")
```
AleCandido, line 101: This is not the most Pythonic way to check if something is iterable: something is iterable if you can iterate it.

So, if you don't care about the specific error, you could just write:
```suggestion
        return type(parameters)(cls[par.upper()] for par in parameters)
```
otherwise, catch the error with `try ... except`, and raise your favorite error.

AleCandido: I'm not sure the constructor would work for any iterable. But if it does, why not :)

And even if it works for the types you're interested in, it's already good enough, and just better even if it works with something else.
This is in the spirit above: you don't check nor ask permission, you do it. And if some operation does not apply, it will raise its own error (this is the duck typing application, and the goal is to make your code cleaner, less checks and errors raised, but more general at the same time).

src/qibosoq/components.py
rodolfocarobene: Yes I saw it :-(

AleCandido (approved):

PR #43, opened 2023-06-21
Split qickprogram file

rodolfocarobene:
Checklist before review:
- [x] Split qick.programs.p[y file into smaller ones in progras folder
- [x] move execute_program from connection handler

Checklist before merge:
- [x] Tests are passing.


src/qibosoq/rfsoc_server.py
```diff
@@ -16,12 +16,76 @@
...
 from qibosoq.components import Config, OperationCode, Pulse, Qubit, Sweeper
-from qibosoq.qick_programs import ExecutePulseSequence, ExecuteSweeps
+from qibosoq.programs.pulse_sequence import ExecutePulseSequence
+from qibosoq.programs.sweepers import ExecuteSweeps
 
 logger = logging.getLogger(cfg.MAIN_LOGGER_NAME)
 qick_logger = logging.getLogger(cfg.PROGRAM_LOGGER_NAME)
 
 
+def execute_program(self, data: dict) -> dict:
```
AleCandido, line 26: Maybe don't call it `self` :)

(it's not a special/reserved word for Python, but it is a bit special for people reading Python)

AleCandido: Moreover, they are now swapped in the call, so you should either swap here:
```suggestion
def execute_program(data: dict, soc) -> dict:
```
or make the call consistent.

doc/source/api-reference/qibosoq.rst
AleCandido, line 1: E.g. Qibocal they are modifying these files. However, to me, it's fine even without any modification at all, possibly even better (to separate generated and human-written content).

If this is fully generated, you could consider ignoring this file in the repo :)
(it can be regenerated identical at any time, and it won't appear in the diffs)

src/qibosoq/rfsoc_server.py
rodolfocarobene: I thought I did it, but I didn't... thanks!

doc/source/api-reference/qibosoq.rst
rodolfocarobene: Wait they can be generated automatically? Both here, in qibolab and qibocal I think they are manually written (and therefore usually outdated)

src/qibosoq/programs/sweepers.py
```diff
@@ -0,0 +1,146 @@
...
+        self.sync_all(self.wait_initialize)
+
+
+SWEEPERS_TYPE = {
+    Parameter.FREQUENCY: "freq",
+    Parameter.AMPLITUDE: "gain",
+    Parameter.BIAS: "gain",
+    Parameter.RELATIVE_PHASE: "phase",
+    Parameter.START: "t",
+}
```
AleCandido, line 146: This you could do it directly with the enum:

```py
class Parameter(Enum):
    """Available parameters for sweepers."""

    FREQUENCY = "freq"
    AMPLITUDE = "gain"
    ...
```
and use it like:
```py
Parameter[sweeper.parameters[0]].value
```

(unless you're already using somewhere that the enum should be a `IntEnum`, but it doesn't seem the case in `main`)

doc/source/api-reference/qibosoq.rst
AleCandido: It can.
I think that @andrea-pasquale explicitly wanted to avoid, because he wanted to mix human written content with docstring extracted one (I usually place this kind of content directly in the source, but you might not like it, because it's adding more lines to your code - even if brought to the extreme, this style has also a [name from D. Knuth]()).

However, if you want to run it manually, there is a CLI tool for doing it:
https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html
if instead you want it to be run every time you generate the docs (if the source has been updated, in a `make`-like fashion), then you could add to your `conf.py`:
https://github.com/readthedocs/readthedocs.org/issues/1139#issuecomment-312626491
(in the end, `conf.py` is a source file that is executed every time you run Sphinx, so you could put whatever in there...)

src/qibosoq/programs/sweepers.py
rodolfocarobene: Simple Enum was giving problem in the serialization (something similar to https://stackoverflow.com/questions/24481852/serialising-an-enum-member-to-json)...

AleCandido: I see, that makes sense. An `IntEnum` is an enum whose variants are also `int`s, so it shouldn't be a problem to serialize.

For the great good and added benefits, in py3.11 they introduced [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum).
However, we're poor men living in a py3.9 world, so we should give up on the added benefits and just do the mixing ourselves:
```py
class Parameter(str, Enum):
    """Available parameters for sweepers."""

    FREQUENCY = "freq"
    AMPLITUDE = "gain"
    ...
```
I tested it serializes well with `json.dumps`, but I leave you the pleasure to test it in Qibosoq as well :)

rodolfocarobene: Nice!

rodolfocarobene:
I changed the type hint from `collections.abc.Iterable` to `typing.Iterable`... Not sure if it was a reason for your proposal of the first, but it was complaining that it was not subscriptable

src/qibosoq/components.py
```diff
@@ -1,9 +1,8 @@
 """Various helper objects."""
 
-from collections.abc import Iterable
 from dataclasses import dataclass, field
 from enum import IntEnum, auto
-from typing import List, Union, overload
+from typing import Iterable, List, Union, overload
```
AleCandido, line 5: It was fine to import `Iterable` from `collections.abc`, since py3.9 it is the recommended way.
(of course, fine unless we have to support py3.8)

AleCandido: In the same spirit you can use `list` instead of `List`

src/qibosoq/programs/sweepers.py
```diff
@@ -1,8 +1,7 @@
 """Program used by qibosoq to execute sweeps."""
 
 import logging
-from collections.abc import Iterable
-from typing import List, Tuple, Union
+from typing import Iterable, List, Tuple, Union
```
AleCandido, line 4: Same of above.

src/qibosoq/components.py
rodolfocarobene: On the board there is python 3.8.2

AleCandido: Ah, ok

rodolfocarobene:
Python 3.8 added to the workflows! Indeed I had never checked which python version I was using on board...

AleCandido:
You could also include/exclude combinations from the matrix:
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations

tests/test_components.py
```diff
@@ -35,4 +35,4 @@ def test_parameter_variants():
 
     assert converted_list == expected
     assert list(converted_tuple) == expected
-    assert list(converted_set) == expected
+    assert list(converted_set).sort() == expected.sort()
```
AleCandido, line 38: This is not doing what you believe is doing: you're comparing `None == None`, because `list.sort()` doesn't return anything:
```suggestion
    assert sorted(converted_set) == sorted(expected)
```

AleCandido: The other option (convenient if `converted_set` is already a `set`) is just to use `set` operations:
```suggestion
    assert len(converted_set ^ set(expected)) == 0
```
https://docs.python.org/3/library/stdtypes.html#frozenset.symmetric_difference

AleCandido (approved):

AleCandido:
As usual, I'd suggest merging first #41 (unless you have an explicit motivation to do otherwise)

rodolfocarobene:
> As usual, I'd suggest merging first #41 (unless you have an explicit motivation to do otherwise)

The motivation would be that #41 is not approved yet :-)

AleCandido:
> The motivation would be that #41 is not approved yet :-)

Then just wait for it ^^. The reason is that this would be merged into that, and the diff would become the sum of the two (slowing down the review even more)

PR #45, opened 2023-06-26
Fix time registers

rodolfocarobene:
Time-Sweepers are not correctly understood by qick since there is a mismatch in name: sometimes they are called "t" registers and sometimes "time" registers. This fixes the problem

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/src/qibosoq/programs/sweepers.py
+++ b/src/qibosoq/programs/sweepers.py
@@ -94,7 +94,7 @@ class ExecuteSweeps(FluxProgram, NDAveragerProgram):
                     max_gain = int(self.soccfg["gens"][gen_ch]["maxv"])
                     starts = (sweeper.starts * max_gain).astype(int)
                     stops = (sweeper.stops * max_gain).astype(int)
-                elif sweeper.parameter[idx] is Parameter.START:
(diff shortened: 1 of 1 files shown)


src/qibosoq/programs/sweepers.py
```diff
@@ -94,6 +94,9 @@ def add_sweep_info(self, sweeper: Sweeper):
                     max_gain = int(self.soccfg["gens"][gen_ch]["maxv"])
                     starts = (sweeper.starts * max_gain).astype(int)
                     stops = (sweeper.stops * max_gain).astype(int)
+                elif sweeper.parameters[idx] is Parameter.START:
+                    # TODO tell qick of the mismatch (t != time)
```
AleCandido, line 98: If I understood correctly the mismatch is on their side, since `self.get_ren_reg` should be a function of `NDAveragerProgram`.

AleCandido (approved):
Assuming that the mismatch is really due to Qick: fine to patch it, but file an issue on their repo, and link it in this PR.

src/qibosoq/programs/sweepers.py
rodolfocarobene: Yes I think so

PR #56, opened 2023-06-29
deploy on tags

### LINKED ISSUE #55: Releasing workflow on tags
In light of a faster iterative workflow, I'd propose to release wheels for each tag, essentially replacing `github.event_name == 'release' && github.event.action == 'published'` in:
https://github.com/qiboteam/qibosoq/blob/af13fbbdfff786cd967ce5ddc862cad4ff4cd93b/.github/workflows/deploy.yml#L23
with `github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')`.

and changing the event:
https://github.com/qiboteam/qibosoq/blob/af13fbbdfff786cd967ce5ddc862cad4ff4cd93b/.github/workflows/deploy.yml#L3-L8
to just:
```
 on: 
   push:
```

rodolfocarobene:
Closes #55. Purely only Alessandro suggestions

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,10 +2,6 @@ name: Deploy
 
 on:
   push:
-  merge_group:
-  release:
-    types:
-      - published
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


scarrazza:
Frankly, I don't understand why you prefer tags instead of releases, when you release you make a tag and this generates a clear release page in github with all diffs and PRs. I would prefer to keep it simple with a releases-only approach, in symmetry with the other repos.

rodolfocarobene:
My only doubt on these "fast" releases/tags is that I will end up doing a new one almost every PR. But I trust your choice and experience on the matter.
Also in the debate between releases and tags, for me it is almost the same... 
Maybe it is nicer to group several PRs in a single new releases, while still publishing on tags more frequently. But again I don't have strong opinions

PR #57, opened 2023-06-29
Documentation update

rodolfocarobene:
Closes #50.
Still work in progress :-)

Checklist before merge:
- [ ] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Documentation is updated.

PR #52, opened 2023-06-29
firing pulse now relative in time

rodolfocarobene:
Closes #51, following the discussion in https://github.com/openquantumhardware/qick/issues/155, I'm moving all the timing from absolute to relative.

Now to play a pulse sequences [A, B, C] we are doing:
```python
pulse(A, t=A.start)
pulse(B, t=B.start)
pulse(C, t=C.start)
```
After this PR it should become:
```python
delay(A.start)
pulse(A)
delay(B.start - A.start)
pulse(B)
delay(B.start - C.start)
pulse(C)
```

this will fix a bug relative to the time sweepers and avoid the "mismatch" in the registers names that we thought existed.
In general this should not be extremely difficult, but I have some doubt that my current implementation will work for multiplexed readout. I will have to check.
Moreover, also a PR in qibolab needs to be opened to take into account this change

Checklist before merge:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Documentation is updated.

PR #58, opened 2023-07-01
Add automated `apidoc`

AleCandido:
Apidoc added

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -13,6 +13,7 @@
 #
 import sys
 from pathlib import Path
+
 from sphinx.ext import apidoc
 
 sys.path.insert(0, str(Path.cwd().absolute()))
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,24 +1,11 @@
 """Configuration file for sphinx documentation."""
-# Configuration file for the Sphinx documentation builder.
-#
-# This file only contains a selection of the most common options. For a full
-# list see the documentation:
-# https://www.sphinx-doc.org/en/master/usage/configuration.html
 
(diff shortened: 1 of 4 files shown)


doc/source/conf.py
```diff
@@ -11,10 +11,12 @@
...
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-import os
 import sys
+from pathlib import Path
 
-sys.path.insert(0, os.path.abspath("."))
+from sphinx.ext import apidoc
+
+sys.path.insert(0, str(Path.cwd().absolute()))
```
AleCandido, line 19: Can we completely remove this line? @rodolfocarobene

AleCandido:
We should tell `pydocstyle` to only have a look inside `src/`...

doc/source/conf.py
rodolfocarobene: Yes definitely, I'm not sure why it is even there, but probably from some careless copy-pasting

rodolfocarobene:
> We should tell `pydocstyle` to only have a look inside `src/`...
> (but I can also provide a minimal docstring for the `conf.py` function, all in all it's not a bad idea)

I think a super-minimal docstring is nice :-)

doc/source/conf.py
AleCandido: I believe that `sphinx-quickstart` generated it. It's just not very useful to us, because we always work with installed environments, even in development mode, so there is no need to manually add the path of the package.

I removed it :)

AleCandido:
As usual, I'd merge this after #52 and #57

rodolfocarobene (approved):

PR #67, opened 2023-07-19
Fixes

rodolfocarobene:
Checklist before merge:
- [ ] Tests are passing.


rodolfocarobene:
Hi, this PR should now be ready for review although it still need some testing with hardware.
Mainly I implemented some corrections for mypy (it would be nice to have it as a workflow as you proposed some time ago). In some occasions I was not able to fix the mypy error and resolved adding a `type: ignore`... maybe there was some better solution.

Also, I moved the average argument in the config object (before it was outside, in the "first scope" of the commands dictionary). I still need to fix documentation

src/qibosoq/programs/base.py
```diff
@@ -274,9 +275,11 @@ def collect_shots(self) -> Tuple[npt.NDArray[np.float64], npt.NDArray[np.float64
                 lengths.append(self.soc.us2cycles(pulse.duration, gen_ch=ro_ch))
             adcs.append(adc_ch)
 
-        adcs, adc_count = np.unique(adcs, return_counts=True)
+        unique_adcs, adc_count = np.unique(adcs, return_counts=True)
 
-        for idx, adc_ch in enumerate(adcs):
+        shape: Tuple  # used for mypy
```
AleCandido, line 280: Convert below in an `if` expression:
```py
shape: Tuple = (count, self.expts, self.reps) if self.expts is not None else (count, self.reps)
```

Also consider that tuples are like structs with anonymous fields, so having the same variable with two different lengths is like making a union, even though the union is just named `Tuple` (that here could also be `tuple`).
However, NumPy accepts the union, so it is consistent. This is just a warning, not a change request.

src/qibosoq/programs/flux.py
```diff
@@ -56,7 +57,7 @@ def set_bias(self, mode: str = "sweetspot"):
                 stdysel="last",
                 freq=0,
                 phase=0,
-                gain=int(max_gain * qubit.bias),
+                gain=int(max_gain * qubit.bias),  # type: ignore
```
AleCandido, line 60: Here there is an explicit cast, so I'm not imagining why you need to disable the type check...

src/qibosoq/programs/pulse_sequence.py
```diff
@@ -2,7 +2,7 @@
 
 import logging
 
-from qick import AveragerProgram
+from qick import AveragerProgram  # type: ignore
```
AleCandido, line 5: Other instance of import check disabling

src/qibosoq/programs/sweepers.py
```diff
@@ -1,11 +1,11 @@
...
 
 import logging
-from typing import Iterable, List, Tuple, Union
+from typing import Iterable, List, Union
 
 import numpy as np
-from qick import NDAveragerProgram, QickSoc
-from qick.averager_program import QickSweep, merge_sweeps
+from qick import NDAveragerProgram, QickSoc  # type: ignore
+from qick.averager_program import QickSweep, merge_sweeps  # type: ignore
```
AleCandido, line 8: If the problem is `qick` not being properly typed, we could whitelist it in the `mypy` configs.

src/qibosoq/programs/sweepers.py
```diff
@@ -46,24 +39,20 @@ def add_sweep_info(self, sweeper: Sweeper):
...
-        sweeper.stops = np.array(sweeper.stops)
+
         if sweeper.parameters[0] is Parameter.BIAS:
             for idx, jdx in enumerate(sweeper.indexes):
                 gen_ch = self.qubits[jdx].dac
                 sweep_type = "gain"
                 std_register = self.get_gen_reg(gen_ch, sweep_type)
                 swept_register = self.new_gen_reg(gen_ch, name=f"sweep_bias_{gen_ch}")
-                self.bias_sweep_registers[gen_ch] = (swept_register, std_register)
+                self.bias_sweep_registers[gen_ch] = (swept_register, std_register)  # type: ignore
```
AleCandido, line 51: To be checked

src/qibosoq/programs/sweepers.py
```diff
@@ -46,24 +39,20 @@ def add_sweep_info(self, sweeper: Sweeper):
...
 
                 max_gain = int(self.soccfg["gens"][gen_ch]["maxv"])
-                starts = (sweeper.starts * max_gain).astype(int)
-                stops = (sweeper.stops * max_gain).astype(int)
+                starts = (np.array(sweeper.starts) * max_gain).astype(int)
```
AleCandido, line 54: Which is the type of `sweeper.starts`?

src/qibosoq/rfsoc_server.py
```diff
@@ -13,7 +8,7 @@
 from socketserver import BaseRequestHandler, TCPServer
 from typing import Dict, List
 
-from qick import QickSoc
+from qick import QickSoc  # type: ignore
```
AleCandido, line 11: Qick again (maybe that's another thing we could do for Sho: typing his Python package)

src/qibosoq/rfsoc_server.py
```diff
@@ -64,7 +59,7 @@ def execute_program(data: dict, qick_soc: QickSoc) -> dict:
     )
 
     asm_prog = program.asm()
-    qick_logger.handlers[0].doRollover()
+    qick_logger.handlers[0].doRollover()  # type: ignore
```
AleCandido, line 62: To be checked (but Qick once more)

src/qibosoq/client.py
```diff
@@ -34,7 +34,7 @@ def connect(server_commands: dict, host: str, port: int) -> Tuple[list, list]:
         results = json.loads(received.decode("utf-8"))
         if isinstance(results, str) and "Error" in results:
             raise QibosoqError(results)
-        return results["i"], results["q"]
+        return results["i"], results["q"]  # type: ignore
```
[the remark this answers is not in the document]
rodolfocarobene: here I suppose we could do something like : `results : Union[Dict, str]`

src/qibosoq/components/base.py
```diff
@@ -49,7 +51,7 @@ class Parameter(str, Enum):
 
     @overload
     @classmethod
-    def variants(cls, parameters: str) -> "Parameter":
+    def variants(cls, parameters: str) -> "Parameter":  # type: ignore
```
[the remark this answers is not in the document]
rodolfocarobene: I receive an error with something like "incompatible return" so I suppose it is the string the problem

src/qibosoq/programs/base.py
```diff
@@ -7,8 +7,8 @@
 
 import numpy as np
 import numpy.typing as npt
-from qick import QickProgram, QickSoc
-from qick.qick_asm import QickRegister
+from qick import QickProgram, QickSoc  # type: ignore
+from qick.qick_asm import QickRegister  # type: ignore
```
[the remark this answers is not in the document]
rodolfocarobene: Otherwise I get an error regarding incomplete imports

src/qibosoq/programs/flux.py
rodolfocarobene: qubit.bias can be None (not here, but in general) so I get an error since I could be multiplying int and None

src/qibosoq/programs/sweepers.py
rodolfocarobene: Ok, it makes sense

src/qibosoq/programs/sweepers.py
rodolfocarobene: it's a list

src/qibosoq/rfsoc_server.py
rodolfocarobene: I think mypy complains because of the import, not really because of qick itself. I added the ignore for every qick import

src/qibosoq/rfsoc_server.py
```diff
@@ -42,7 +37,7 @@ def execute_program(data: dict, qick_soc: QickSoc) -> dict:
         (dict): dictionary with two keys (i, q) to lists of values
     """
     opcode = OperationCode(data["operation_code"])
-    args = ()
+    args = (_ for _ in ())  # initialize empty generator for sequences
```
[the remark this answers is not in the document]
rodolfocarobene: type(()) is tuple, but for sweepers is a generator and mypy complains. In this way is an empty generator

src/qibosoq/rfsoc_server.py
rodolfocarobene: No this is not qick, is something related to the loggers that I define. 
`qick_logger.handlers` is typed as a list of `Handler`, but `doRollover` is a function just of `RotatingHandler`

rodolfocarobene:
Honestly some of the corrections that I added make me doubtful using mypy is the greatest idea...

src/qibosoq/components/base.py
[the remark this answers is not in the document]
AleCandido: I will investigate

src/qibosoq/programs/flux.py
AleCandido: Instead of ignoring the error, if this property is lost by the type you receive (because you generate a type that is not `None`, but then you only declare the broader class) assert that the property still holds.

```py
assert qubit.bias is not None
```

The best would be to be able to do it statically, such that at linting time you're already sure that it won't be `None`, since it's something that you know to be true. However, if you're not able to state statically, at least tell it dynamically (such that sooner or later we would get an error in case of violation, either from `mypy` or from Python itself).

src/qibosoq/programs/sweepers.py
AleCandido: Why not an `np.array` in the first place? Is there a case in which you want to `.append()` or something like that?

src/qibosoq/rfsoc_server.py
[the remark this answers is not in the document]
AleCandido: Can you point me where the sweepers' counterpart is defined?

src/qibosoq/rfsoc_server.py
AleCandido: Ah, you're right. I read `qick`, but I have been superficial. Where is that `.handlers` gets its `list[Handler]` type? Standard library should be properly typed since a while...

AleCandido:
> Honestly some of the corrections that I added make me doubtful using mypy is the greatest idea...

I know the feeling: it's cumbersome. But it's like that in the first place because the program grew without paying too much attention to types, so a bunch of things are working at runtime because they are working, even though `mypy` couldn't understand it.

However, what `mypy` can't understand, even if it's working now, can unexpectedly stop working in the future, because constraints that now are respected might be violated by unaware developers.

All in all, it's not something I have imposed yet, exactly because it's cumbersome at the beginning. This is the reason why there is no workflow. However, if we really manage to merge this PR without `# type: ignore` comments, I would actually try to start using such a workflow with Qibosoq. It could really help to avoid a bunch of errors in advance.
It might be annoying since it's pedantic (wrt standard run-time-only Python), like Pylint, but it's eventually paying, and at some point we'll do it properly automatically.

src/qibosoq/client.py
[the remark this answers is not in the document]
AleCandido: But a string can only be accessed with integers, not with other strings. How would it be possible at runtime to still have a string here?

The error most likely survives because you're checking with `and` above, but maybe you'd need more something like:
```py
if isinstance(results, str):
    if "Error" in results:
        raise QibosoqError(results)
    raise ValueError("something")
```

I know it's Pythonic to fail if an operation is not possible. But what is Pythonic actually change a bit when you start using types (it's quite uncontroversial that types, even with inference, are a different paradigm from the run-time-only Python).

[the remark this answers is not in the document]
rodolfocarobene: This indeed works, but then I would simply remove the second if:
```python
if isinstance(results, str):
    raise QibosoqError(results)
```
It still makes enough sense right?

src/qibosoq/rfsoc_server.py
AleCandido: Yes, but the `import` statement has no problem on its own. Let me check the exact error...

src/qibosoq/programs/flux.py
rodolfocarobene: some line up I have an if to break the for... I kinda like the idea and it seems already enough
I thought of modifying it like this:
```python
if qubit.bias is None or flux_ch is None or qubit.bias == 0:
                continue
```
what do you think?

rodolfocarobene: Or even adding a property like "qubit.fluxed"

AleCandido: Both options should be fine, with the second being good if you want to add extra docs explanations.

rodolfocarobene: Mypy does not seem to like the property... still same error. I have now resorted to the if

src/qibosoq/programs/base.py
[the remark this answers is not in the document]
rodolfocarobene:
```
src/qibosoq/programs/sweepers.py:7: error: Skipping analyzing "qick": module is installed, but missing library stubs or py.typed marker  [import]
src/qibosoq/programs/sweepers.py:8: error: Skipping analyzing "qick.averager_program": module is installed, but missing library stubs or py.typed marker  [import]
```
I can add the mypy config file excluding the qick import

src/qibosoq/programs/sweepers.py
rodolfocarobene: It was because the communication protocol is "based" on json and ndarray are not natively serializable... Clearly I can convert them to list before communication. The two options seem pretty equivalent to me

src/qibosoq/rfsoc_server.py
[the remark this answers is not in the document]
rodolfocarobene: Here it is:
https://github.com/qiboteam/qibosoq/blob/d0bd2e2c63862df4e0736a6ca3542cf73ecd9f74/src/qibosoq/rfsoc_server.py#L49

Maybe I can use a tuple here, to avoid the weirdness of the empty generator

src/qibosoq/rfsoc_server.py
rodolfocarobene: Actually, I think it's mypy that infers that typing. In the logging module it seems non typed:
https://github.com/python/cpython/blob/a6313d78f21f79ca64dedd38e637509dc530a1b6/Lib/logging/__init__.py#L1456

src/qibosoq/components/base.py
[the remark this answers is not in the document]
rodolfocarobene: The complete error is:
```
src/qibosoq/components/base.py:54: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]
```
if I swap the two overloaded function I get:
```
src/qibosoq/components/base.py:59: error: Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same or broader  [misc]
```
So I think the problem is that a `str` is also a `Iterable[str]`...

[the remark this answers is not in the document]
rodolfocarobene: Meh: https://stackoverflow.com/questions/71656644/python-type-hint-for-iterablestr-that-isnt-str

src/qibosoq/rfsoc_server.py
AleCandido: https://github.com/python/typeshed/blob/main/stdlib/logging/handlers.pyi

rodolfocarobene: So here it is! https://github.com/python/typeshed/blob/754473a30d50b918ac6f1a824f41600806defa94/stdlib/logging/__init__.pyi#L105

src/qibosoq/programs/sweepers.py
AleCandido: Ok, you're right.

I believe we should just pick a scheme: converting immediately, or as late as possible.

This of course is only possible when data have a "true format" and a different representation. In which case, I would argue that the *as soon as possible* scheme is preferable, since you already know you're going to consume data in that fashion, and you'd avoid doing it multiple times.
The advantage of doing it late would be to never do it when it's not strictly needed (but I believe this is more rarely happening).

So, I would directly deserialize into an array, when you load the JSON representation into the object (if possible).

src/qibosoq/rfsoc_server.py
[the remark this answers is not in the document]
rodolfocarobene: Having added a function to convert sweepers from dict to sweeper objects with arrays, this is just a a list so it should be fine simply with
```python
args = []
```

src/qibosoq/components/base.py
```diff
@@ -75,9 +77,9 @@ class Sweeper:
     """Number of points of the sweeper."""
     parameters: List[Parameter]
     """List of parameter to update."""
-    starts: List[Union[int, float]]
+    starts: Union[List[float], np.ndarray]
```
AleCandido, line 80: Isn't it possible to support only `npt.NDArray`? (dropping lists)

rodolfocarobene: Maybe yes, but at the cost of more complexity in the serialization/deserialization for the communication...
I could add a dict_factory to use the asdict to convert arrays into list and then change the __init__ to accept also lists for the serialization.
This will also change the "logic" of the commands dictionary that will not be anymore a strict dictionary representation of the objects...

Not sure if it's much better than now

AleCandido: > Maybe yes, but at the cost of more complexity in the serialization/deserialization for the communication...

More or less that was the proposal above: relegate all the complexity to the serialization/deserialization, and after that you could rely on them being arrays, avoid any further conversion.

So, the idea is that when you have the object, it contains arrays (which are other objects), and when you serialize, the object becomes a dict, and the arrays lists.

rodolfocarobene: With these lines I am sure that they are arrays everywhere on the server:

AleCandido: On the client side, I would actually construct either everything in the serialized representation, or everything in the object fashion (using `qibosoq` classes and methods).

So, I'd clearly separate the two worlds: serialized and deserialized. And then, make tools to make the two of them equivalent.
What I wouldn't do is to blend the two worlds, because then you add further conversions, you need to take into account more cases, and sometimes you also need to take manual steps before converting *to the other side* (i.e. if you are in the middle, but you only have a map from one edge to the other and conversely, you might have to move to the closest edge before being able to jump to the other).

rodolfocarobene: I have now removed the support for lists in starts and stops, I left it for indexes and parameters...
Should I also remove it from them? 
Actually I tried, but had some problem with the numpy.array of Parameter objects.  
It shouldn't matter much anyway, I suppose...

AleCandido: I exposed above the general principle, but do not worry too much about getting everything perfect in just one shot.

The best fix in this case, if everything still works, and only Mypy is complaining, just make the workflow succeed anyhow, ignoring the outcome of Mypy.
Whenever we will solve all Mypy problems, we could drop the toggle and allow workflow failures.

rodolfocarobene: Actually the problem was not only of mypy, but things were not working at all... But I think that, as it is now, it's good enough :-)

rodolfocarobene:
Hi @AleCandido, this should be ready (and also the mypy one), I would try to merge as soon as posible so that I can focus on the flux pulses one (#29) and on the paper.

Things I added in the last commits:
* sweeper post_init so that even if you initilize them with lists, they are working with arrays (and the serialization does not brake)
* updated documentation
* log folder gets created if it doesn't exsist (on linux)

AleCandido:
> Things I added in the last commits:
> 
>     * sweeper post_init so that even if you initilize them with lists, they are working with arrays (and the serialization does not brake)
> 
>     * updated documentation
> 
>     * log folder gets created if it doesn't exsist (on linux)

Thanks for the recap, I'll try to review again as soon as possible.

In case you didn't know, GitHub offers a similar feature:
image
(of course it can not summarize, for that I usually read the commit messages)

doc/source/getting-started/communication.rst
```diff
@@ -4,6 +4,8 @@ Communication protocol
 Originally conceived as a component of ``Qibolab``, ``Qibosoq``, has since evolved into an independent entity, no longer reliant on ``Qibolab`` for its functionality.
 Due to is standalone nature, it can now be utilized by any application that adheres to the ``Qibosoq`` communication protocol.
 
+For the user, leveraging the ``qibosoc.client`` module, it is not strictly required to know the communication protocol.
```
AleCandido, line 7:
```suggestion
For the user, leveraging the ``qibosoq.client`` module, it is not strictly required to know the communication protocol.
```

README.md
```diff
@@ -8,23 +8,36 @@ Repository for developing server side of RFSoC fpga boards
...
 
-The complete documentation for can be found at [qibo.science/qibosoq/stable](https://qibo.science/qibosoq/stable/)
+The complete documentation for can be found at:
 
-## Installation
+* [qibo.science/qibosoq/stable](https://qibo.science/qibosoq/stable/)
+* [qibo.science/qibosoq/latest](https://qibo.science/qibosoq/latest/)
+
+
+## Installation from pip
```
AleCandido, line 17: It's nice to have installation instructions on the README, but this is duplicating the information in the docs, so both have to be maintained consistent.

To simplify, I'd consider adding just a link to the docs. What do you think?

src/qibosoq/client.py
```diff
@@ -32,22 +34,35 @@ def connect(server_commands: dict, host: str, port: int) -> Tuple[list, list]:
                 break
             received.extend(tmp)
         results = json.loads(received.decode("utf-8"))
-        if isinstance(results, str) and "Error" in results:
+        if isinstance(results, str):
             raise QibosoqError(results)
         return results["i"], results["q"]
 
 
+def serialize_sweeper(sweeper: Sweeper) -> dict:
```
AleCandido, line 42: Since this is a function of a single argument, i.e. the Qibosoq class `Sweeper`, this would work perfectly as a `Sweeper` property, e.g. `.dict`/`.serialized`/`.raw`.

The main advantage is that you avoid the need for a further import.

README.md
rodolfocarobene: Seems reasonable, thanks!

src/qibosoq/log.py
```diff
@@ -3,13 +3,20 @@
...
+import sys
 from typing import Tuple
 
 import qibosoq.configuration as cfg
 
 
 def configure_logger(name: str, filename: str, backup_count: int):
     """Create and configure logger."""
+    # if the log directory does not exsist, create it
+    if sys.platform == "linux":
```
AleCandido, line 15: Why did you restric to `linux` only?

src/qibosoq/log.py
```diff
@@ -3,13 +3,20 @@
...
+        dir_path = "/".join(filename.split("/")[:-1])
+        if not os.path.exists(dir_path):
+            os.mkdir(dir_path)
```
AleCandido, line 18: You can convert all of this to `pathlib`:

```suggestion
        dir_path = Path(filename).parent
        if dir_path.exists():
            dir_path.mkdir()
```

rodolfocarobene: This is why I was restring things to linux, because of the '/', but with pathlib is nicer.
Anyway, this will always run only on the server that is linux

src/qibosoq/programs/base.py
```diff
@@ -244,23 +239,23 @@ def acquire(
         # if there are no readouts, temporaray set 1 so that qick can execute properly
         reads_per_rep = 1 if readouts_per_experiment == 0 else readouts_per_experiment
 
-        res = super().acquire(
+        res = self.acquire(  # pylint: disable=E1123
```
AleCandido, line 242: Actually, Pylint seems right in flagging this as an error, why did you disable?

https://github.com/openquantumhardware/qick/blob/a17f2dbc449bfcc723da33795bac3d4fc11cba25/qick_lib/qick/qick_asm.py#L1454

src/qibosoq/log.py
rodolfocarobene: fixed

src/qibosoq/rfsoc_server.py
```diff
@@ -127,7 +133,7 @@ def handle(self):
         try:
             data = self.receive_command()
             results = execute_program(data, self.server.qick_soc)
-        except Exception as exception:  # pylint: disable=bare-except, broad-exception-caught
+        except Exception as exception:  # pylint: disable=W0612,W0718
```
AleCandido, line 136: Btw, you're giving a name to the `exception`, but then you're not using it (I guess because `traceback` already knows about it).

Maybe you could drop the `as exception` part.

src/qibosoq/programs/base.py
rodolfocarobene: Pylint see an error because it thinks that QickProgram.acquire will be called, actually we always call AveragerProgram.acquire (or NDAverager) that has the readouts_per_experiment agument

https://github.com/openquantumhardware/qick/blob/a17f2dbc449bfcc723da33795bac3d4fc11cba25/qick_lib/qick/averager_program.py#L71

src/qibosoq/rfsoc_server.py
rodolfocarobene: Yes I think I can remove it... I was probably following some reference on how to log exceptions (following it blindly)

src/qibosoq/programs/base.py
AleCandido: I actually thought the same as Pylint: then why do we inherit from `QickProgram`, and not from `AveragerProgram`?

rodolfocarobene: Because the base program is used both by the pulse sequence and the sweeper program and they are, respectively, a AveragerProgram and a NDAveragerProgram. Both inheriting from QickProgram

rodolfocarobene: This issue is just caused by the fact that there is this change of name in the function parameters. If this was not the case, I think this would have been perfectly fine... because of that I have added the pylint exception

rodolfocarobene: > (it would be even interesting to know why they have a different and incompatible interface...)
My guess: to be completely sure of which function they are actually using...

AleCandido: Maybe it's worth to open a tiny issue on Qick.

If they are inheriting, and the base program is establishing an interface, it would be nice to preserve it.

Actually, they could have used different methods names (to signal that they are different operations). And this would have been fine from their side.
In which case we could not use just `self.acquire`, and we might have had to distinguish between `Averager` and `NDAverager`.

What we could do is to implement an `BaseProgram._acquire`, that does everything but the call to the underlying `acquire` function, and implement this in the subclasses

src/qibosoq/programs/base.py
```diff
@@ -223,15 +222,11 @@ def body(self):
 
         self.wait_all()
 
-    # pylint: disable=unexpected-keyword-arg, arguments-renamed
-    def acquire(
+    def perform_experiment(
```
AleCandido, line 225: Since you changed the name, maybe you should take the chance to also update the docstring with a more explicit one.

src/qibosoq/programs/base.py
```diff
@@ -244,23 +239,23 @@ def acquire(
...
         reads_per_rep = 1 if readouts_per_experiment == 0 else readouts_per_experiment
 
-        res = super().acquire(
+        res = self.acquire(  # pylint: disable=E1123
             soc,
             readouts_per_experiment=reads_per_rep,
-            load_pulses=load_pulses,
-            progress=progress,
-            debug=debug,
+            load_pulses=True,
```
AleCandido, line 245: Just to understand myself: why are you hardcodingb these now?

It's a perfectly reasonable choice, and might save boilerplate, but I wanted to understand why now and why those values.

src/qibosoq/programs/base.py
```diff
@@ -244,23 +239,23 @@ def acquire(
...
+            progress=False,
+            debug=False,
         )
         # if there are no actual readouts, return empty lists
         if readouts_per_experiment == 0:
             return [], []
         if average:
             # for sweeps res has 3 parameters, the first is not used
-            return res[-2:]
+            return np.array(res[-2]).tolist(), np.array(res[-1]).tolist()
```
AleCandido, line 254: Are you sure you still need the double conversion?

This should be the result of a union (since `np.array` is normalizing both to an array, and then you turn it into a list).

src/qibosoq/programs/base.py
```diff
@@ -318,7 +310,7 @@ def declare_gen_mux_ro(self):
             ro_ch=adc_ch_added[0],
         )
 
-    def add_muxed_readout_to_register(self, ro_pulses: List[Pulse]):
+    def add_muxed_readout_to_register(self, ro_pulses: List[Rectangular]):
```
AleCandido, line 313: Again, just to understand myself: do you know why multiplexing is limited to rectangular pulses?

src/qibosoq/programs/flux.py
```diff
@@ -84,3 +86,16 @@ def body(self):
         self.set_bias("zero")
         self.soc.reset_gens()
         self.sync_all(self.relax_delay)
+
+    def declare_zones_and_ro(self, sequence: List[Pulse]):
+        """Declare all nqz zones and readout frequencies.
```
AleCandido, line 91: Just for myself: what it an _nqz zone_?

And, btw, why it's declaring all qubit channels but feedback?

src/qibosoq/programs/base.py
rodolfocarobene: They were already hardcoded in the rfsoc_server.py, I just moved them here. In any case, these are the default values, I am writing them explicitly for clarity and stability (if something changes in qick) but I could also remove them

src/qibosoq/programs/flux.py
```diff
@@ -84,3 +86,16 @@ def body(self):
...
+
+        Declares drives, fluxes and readout (mux or not) and readout freq.
+        """
+        self.declare_nqz_zones([pulse for pulse in sequence if pulse.type == "drive"])
+        self.declare_nqz_flux()
```
AleCandido, line 96: And how does it work this empty declaration?
(and the following as well)

Since it's taking no arguments and returns nothing, this should modify somehow `self`, but I wonder what is doing...

src/qibosoq/programs/base.py
rodolfocarobene: If I remember correctly this is a limitation of qick (I don;t know if software or firmware). 
I think the "set_pulse_register" function, where you set which waveform to use, doesn't work completly... But I would have to check

src/qibosoq/programs/flux.py
rodolfocarobene: Nyquist zone (defined as intervals of fs/2, where fs is the sampling frequency). So basically from 0 to fs/2 is the first and fs/2 to fs is the second and so on.
This declaration is needed because for the first zone the synthesis works differently than for the other zones.

I am declaring it for all the generator channels, the adc does not need it (but needs the declaration of frequency for the downsampling)

src/qibosoq/programs/flux.py
rodolfocarobene: It considers the self.qubits attribute and sees which qubits are biased.
In #29 , this declaration is also extended to flux pulses

src/qibosoq/programs/base.py
rodolfocarobene: think I was getting some errors that could be explained with res[-2] and res[-1] being list of arrays... I didn't investigate much, but it got fixed like this

src/qibosoq/programs/base.py
rodolfocarobene: > Maybe it's worth to open a tiny issue on Qick.

At the moment qibosoq is not compatible with the newest versions of qick, so I had avoided it... I don't think it's really worth it honestly, but sure I may be wrong

> What we could do is to implement an `BaseProgram._perform_experiment`, that does everything but the call to the underlying `acquire` function, and implement this in the subclasses.

I have though of that, but in any case I will end up complicating the code and duplicating some parts

AleCandido: > I have though of that, but in any case I will end up complicating the code and duplicating some parts

It's not urgent, but we could solve this problem, e.g. with a generator (in the general case, you can use a generator, i.e. a function that `yield`s, also to pause the execution and *receive* a value, on top of giving back one to the caller).

I believe we could solve in an elegant enough way, without duplication. But it's definitely not urgent, so please, turn this into an issue (if it will age too much, and we won't care, we'll then close the issue).

src/qibosoq/programs/base.py
AleCandido: Ok, this makes sense, I didn't consider. If they are lists of arrays, `np.array` is sufficiently powerful/flexible to convert them into arrays, and then `.tolist()` is converting them back into nested lists.

But are the lists of arrays coming from Qick?

src/qibosoq/programs/base.py
AleCandido: Thanks for the answer, no current need for a deeper investigation

src/qibosoq/programs/flux.py
rodolfocarobene: What do you mean?
 Both the qubits and the program exist only in qibosoq... I can move the function call somewhere else, but in any case I will have to write it

src/qibosoq/programs/base.py
rodolfocarobene: Yes, I was now looking at the qick source code and I am not sure to find the problem... I think that generally Qick returns arrays of arrays, but in some very specific cases does not follow the rule.
Again, I don't remember for which specific experiment I had to implement this fix, looking at the code it could be something relted to `acquire_decimated` that is used to acquire the non-averaged values

rodolfocarobene:
Is this ready @AleCandido?

.github/codecov.yml
AleCandido, line 1: Not for this PR (not a requirement to merge), but since you introduced the file, I would remove the Codecov comments (like in Qibolab).

https://github.com/qiboteam/qibolab/blob/d5b080487f80da0561e5318a7ccabbd34e5fcf6d/.github/codecov.yml#L7-L8

src/qibosoq/components/base.py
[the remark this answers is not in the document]
AleCandido: This problem is complex, because intrinsic to Python native types (as you found out). Let's forget for the time being.

src/qibosoq/programs/base.py
AleCandido: Ok, let's classify this as a Qick issue. We could investigate later on (if ever).

AleCandido (approved):
For this PR it is more than enough, let's merge :)

Thanks @rodolfocarobene

.github/codecov.yml
```diff
@@ -0,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        threshold: 0.5%
```
AleCandido, line 5: Also consider to slightly increase this threshold: Qibosoq is not as big as Qibolab, so it is rather normal that a PR affects a larger portion of it.

PR #69, opened 2023-07-24
Mypy workflow

rodolfocarobene:
In this PR I added the mypy workflow.

For now, I am performing the check on the usual matrix of os and python version, but maybe we could limit it to a single os... being a static checker I don't think it makes any difference

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,2 +1,0 @@
-[mypy-qick.*]
-ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,2 @@
+[mypy-qick.*]
+ignore_missing_imports = True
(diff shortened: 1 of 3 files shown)


mypy.ini
AleCandido, line 1: Use `pyproject.toml` instead of this.

https://mypy.readthedocs.io/en/stable/config_file.html?highlight=pyproject.toml#using-a-pyproject-toml-file

AleCandido (approved):
Just a tiny comment, but the PR mainly consists of the action and it is fine.

AleCandido:
After this, I'd actually collect in a single action Mypy and Pylint. But at that point I'd write it as reusable, and move to the `workflows` repository.

(that's the goal of the `analysis` dependency group)

PR #85, opened 2023-09-24
Paper runcards

rodolfocarobene:
I added a new folder with the runcards used for the paper, in this way we can link directly the experiments for reproducibility.

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/extras/qibosoq_paper_runcards/README.MD
+++ b/extras/qibosoq_paper_runcards/README.MD
@@ -1,0 +1,18 @@
+# Qibosoq paper runcards
+
+In this folder, we have stored the runcards used for the experiments presented in the Qibosoq paper.
+
+For the timing experiments, we used the following runcards:
+- `routines_benchmarks.yml`
+- `scaling_benchmarks.yml`
+
+These experiments were conducted on a platform analogous to the one described in `my_platform.yml`.
+On the other hand, please note that the calibration experiments can only be exactly reproduced with the same hardware; therefore, the platform runcard used for those experiments is not provided.
+
+These runcards are compatible and were executed with the following software versions:
+
+| Software  | Version |
+|-----------|---------|
+| Qibosoq   | 0.0.4   |
+| Qibolab   | 0.1.0   |
+| Qibocal   | 0.0.3   |
--- a/extras/qibosoq_paper_runcards/scaling_benchmarks.yml
+++ b/extras/qibosoq_paper_runcards/scaling_benchmarks.yml
@@ -1,0 +1,292 @@
+# executed with `qq auto scaling_benchmarks.yml --no-update -o output_folder`
+# all the different sweepers were executed separately
+platform: my_platform  # set to platform runcard
+
+# nshots : 4096 in platform runcard
+
+# 1D Sweepers
+# RO Frequency
+
+  # - id: resonator spectroscopy 1
+  #   # ideal 0.0184
+  #   priority: 00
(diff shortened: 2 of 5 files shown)


extras/qibosoq_paper_runcards/scaling_benchmarks.yml
```diff
@@ -0,0 +1,292 @@
+# executed with `qq auto scaling_benchmarks.yml --no-update -o output_folder`
+# all the different sweepers were executed separately
```
AleCandido, line 2: I know that there is the common habit of uncommenting a piece at a time and run, but this is a bit confusing.

I'd rather suggest splitting this runcard in the 6 sections is consisting of, and keep only descriptive comments (and no commented code at all).

AleCandido: If you want to group them together, a proper way would be a subfolder (but it is not required, choose your favorite way).

extras/qibosoq_paper_runcards/README.MD
```diff
@@ -0,0 +1,18 @@
+# Qibosoq paper runcards
+
+In this folder, we have stored the runcards used for the experiments presented in the Qibosoq paper.
+
+For the timing experiments, we used the following runcards:
+- `routines_benchmarks.yml`
+- `scaling_benchmarks.yml`
+
+These experiments were conducted on a platform analogous to the one described in `my_platform.yml`.
+On the other hand, please note that the calibration experiments can only be exactly reproduced with the same hardware; therefore, the platform runcard used for those experiments is not provided.
```
AleCandido, line 10: I'm not sure this is the best strategy: even though it is unstable, and not working outside the TII, having even that runcard could in principle help to understand what has been done.

In practice, I agree with you that it will change little to nothing.
But if anyone happens to arrive until here, a useless runcard looks better than an omitted one (but maybe this is just a personal perspective).

extras/qibosoq_paper_runcards/scaling_benchmarks.yml
rodolfocarobene: With --no-update it should work even in the same runcard. I removed all the comments and it should now work fine

AleCandido:
`--no-update` should be available as a runcard key (even top-level). If that works, it should either be specified to use it in some note (e.g. `README.md`), or by plugging that key in the runcard itself.

andrea-pasquale:
> `--no-update` should be available as a runcard key (even top-level). If that works, it should either be specified to use it in some note (e.g. `README.md`), or by plugging that key in the runcard itself.

I think that you can simply but `update: False` at the top of the runcard and it should work.

AleCandido:
> I think that you can simply but `update: False` at the top of the runcard and it should work.

The *If that works* part meant something like *if you really do not need to update between subsequent steps*.

andrea-pasquale:
> The _If that works_ part meant something like _if you really do not need to update between subsequent steps_.

I think the only issue could be that if you have a completed uncalibrated chip and you run all the routines sequentially you might end up with values that don't make sense, such as amplitude > 1, which will trigger and error in qibolab therefore stopping the execution.
In this sense, if you care only about performance setting `update=False` it is the safer option.

rodolfocarobene:
Thanks @AleCandido @andrea-pasquale, now that is fixed!

AleCandido (approved):
Thanks @rodolfocarobene

PR #89, opened 2023-10-28
Decouple measurement and pulses

rodolfocarobene:
The idea is to allow for the possibility to perform measurements without pulses.
I've tried to maintain backwards compatibility in order to not break also the qibolab driver. 
Tests are passing locally, but I still have to perform some tests on board.

Checklist before review:
- [x] Tests on hardware
- [x] Fix coverage adding tests
- [x] Add element to the documentation

Checklist before merge:
- [x] Tests are passing.


rodolfocarobene:
This is also working, I'll write here a simple test that can be done performed using a loopback.

The idea is to send a drive pulse to be acquired with a readout pulse with 0 amplitude and with the new "Measurement Pulse" (that does not actually have a pulse). If the two plots are identical this should be a sign that everything is working properly.

```python
from qibosoq.client import execute
from qibosoq.components.base import (
    Qubit,
    OperationCode,
    Config
)
from qibosoq.components.pulses import Rectangular, Measurement, Gaussian

import matplotlib.pyplot as plt
import numpy as np

HOST = "172.18.2.52"
PORT = 6000

pulse_1 = Gaussian(
            frequency = 5400, #MHz
            rel_sigma = 1,
            amplitude = 0.05,
            relative_phase = 0,
            start_delay = 0,
            duration = 0.5,
            name = "drive_pulse",
            type = "drive",
            dac = 6,
            adc = None
)

pulse_2 = Rectangular(
            frequency = 5400, #MHz
            amplitude = 0.0,
            relative_phase = 0,
            start_delay = 0.0,
            duration = 2,
            name = "readout_pulse",
            type = "readout",
            dac = 5,
            adc = 0
)

pulse_3 = Measurement(
    frequency = 5400,
    start_delay = 0.0,
    duration = 2,
    adc = 0,
    dac = 5,
    type = "readout",
)

sequence = [pulse_1, pulse_2]
config = Config(adc_trig_offset=0)
qubit = Qubit()

server_commands = {
    "operation_code": OperationCode.EXECUTE_PULSE_SEQUENCE_RAW,
    "cfg": config,
    "sequence": sequence,
    "qubits": [qubit],
}

i, q = execute(server_commands, HOST, PORT)
values_readout_pulse = np.array(i[0][0]) + 1j * np.array(q[0][0])

sequence[1] = pulse_3
server_commands = {
    "operation_code": OperationCode.EXECUTE_PULSE_SEQUENCE_RAW,
    "cfg": config,
    "sequence": sequence,
    "qubits": [qubit],
}

i, q = execute(server_commands, HOST, PORT)
values_measurement = np.array(i[0][0]) + 1j * np.array(q[0][0])

plt.plot(np.abs(values_readout_pulse))
plt.plot(np.abs(values_measurement))
```

immagine

AleCandido:
Just to understand: what is the use case of the isolated measurement?

rodolfocarobene:
> Just to understand: what is the use case of the decoupled measurement?

The main idea in my mind is to provide more flexibility offering a "new feature".
Then I can see some possible applications like acquiring an arbitrary function directly in loopback from a dac (something that is not directly possible now, without also crating a readout pulse) or waiting for an expected signal (not particularly useful for quantum computing) not generated from the board itself

AleCandido (approved):
The changes should be backward compatible, so in this sense it is perfectly fine.

The part I'm not completely convinced of is if we need this feature, or whether we need it now, and that its inclusion increases a bit the complexity of the code.

However, it's just a minor concern, and it is a concern only on a large scale. It won't be a single new feature to make Qibosoq much more complex, so feel free to merge it :)

PR #93, opened 2023-12-04
Flat top pulse shape

rodolfocarobene:
Checklist before review:
- [x] Test with rfsoc

Checklist before merge:
- [x] Tests are passing.


rodolfocarobene:
I'm finally able to resume testing with RFSoCs! I will work on closing all the issues / open pull requests.

For now, I've tested this new pulse and it works:


immagine

AleCandido (approved):

AleCandido:
@rodolfocarobene in 
https://github.com/qiboteam/qibosoq/blob/9eb405e8d2f2423929560512f1413bbb547c6add/src/qibosoq/programs/base.py#L102

There is a growing "switch" (now with 5 elements), where there is a quite repeating pattern:
- branch on a pulse shape
- do some computation
- call `self.set_pulse_registers()`

The *funny* bit is that the last three elements of the switch (`Gaussian`, `Drag`, and `Arbitrary`) are sharing the same final call, while the first two have their own (in which the main difference is the `style`, set to `"arb"` for the last three).

Could we reconcile all of them to a single call? E.g. by setting a `style` variable.

If the "do some computation" part were massive, I would have suggested splitting in separate functions. However, it seems there is no need, so I would just factorize the call for the time being.

src/qibosoq/programs/flux.py
```diff
@@ -94,8 +94,11 @@ def execute_flux_pulse(self, pulse: Pulse):
             i_vals = np.full(duration, amp)
         elif isinstance(pulse, FluxExponential):
             i_vals = pulse.i_values(duration, max_gain)
+        elif isinstance(pulse, Arbitrary):
+            i_vals = np.array(pulse.i_values)
+            logger.info("Arbitrary shaped flux pulse. q_vals will be ignored.")
         else:
-            raise NotImplementedError("Only Rectangular and FluxExponential flux pulses are supported")
+            raise NotImplementedError("Only Rectangular, FluxExponential and Arbitrary are supported for flux pulses")
```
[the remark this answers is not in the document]
rodolfocarobene: I would have to add another "switch" that could be a bit complex (since for every pulse I should get I-values).
This feature would also probably never be used...

rodolfocarobene:
> If the "do some computation" part were massive, I would have suggested splitting in separate functions. However, it seems there is no need, so I would just factorize the call for the time being.

What do you think of it now? I tried to improve it

src/qibosoq/components/pulses.py
```diff
@@ -32,13 +33,25 @@ class Pulse:
     adc: int
     """ADC to acquire pulse back, for readout pulses."""
 
+    def get_waveform_name(self) -> Optional[str]:
+        """Return waveform name from parameters."""
+        return None
+
+    def get_waveform_style(self) -> Optional[str]:
+        """Return waveform style from parameters."""
+        return "arb"
```
AleCandido, line 42: I know that you want to have methods, since below you want to make use of attributes, but you could at least make these properties
```suggestion
    @property
    def waveform_name(self) -> Optional[str]:
        """Return waveform name from parameters."""
        return None

    @property
    def waveform_style(self) -> Optional[str]:
        """Return waveform style from parameters."""
        return "arb"
```

AleCandido: Moreover: how is `.waveform_name` different from `.name`?

For `.waveform_style` you could also just use `.style`, since it's unambiguous. But `.name` is not, and I'm not sure I understand the difference...

AleCandido (commented):
Thanks for the update, it was exactly what I meant :)

src/qibosoq/components/pulses.py
rodolfocarobene: I've changed them to properties.

"name" is, for the moment, something related to the pulse object and not the waveform... 
To be fair, it is not very useful as a parameter, but it can be nice to have when coding (name could be "drive_pulse", "higher_pulse" and so on) to have qualitative descriptions.
"waveform_name" is a signature of the waveform and depends only on other parameters (so it is also useless to send from client to server)

AleCandido: Ok, so `.name` is just a description (an arbitrary label).

rodolfocarobene: Yes exactly

PR #95, opened 2023-12-13
Fix documentation

### LINKED ISSUE #91: Test documentation
Every code example in the documentation should be properly tested

rodolfocarobene:
Closes #91.

Checklist before review:
- [x] Test Qibosoq examples
- [x] Test Qibolab examples
- [x] Test Qibocal examples

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/doc/source/getting-started/installation.rst
+++ b/doc/source/getting-started/installation.rst
@@ -39,5 +39,5 @@ Now the installation can be done in normal mode or in developer mode (using ``po
     For running qibosoq on board, sudo privileges are required, so the installation must be done with:
 .. code-block:: bash
 
-    sudo -i python -m pip install <path_to_qibosoq>  # normal mode
-    sudo -i python -m poetry install <path_to_qibosoq> # developer mode
+    sudo -E python -m pip install <path_to_qibosoq>  # normal mode
+    sudo -E python -m poetry install <path_to_qibosoq> # developer mode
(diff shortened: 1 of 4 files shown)


AleCandido:
Nice that you found some errors and fixed them!

However, here you're only correcting some snippets, so I guess you just tested them *manually*.
This is already something, and better than broken examples. But consider to test them in the CI (it should be possible using a "dummy" platform, but you will need a dummy mode for Qibosoq itself).

What I mean is to use doctests, as it happens in Qibo itself
https://github.com/qiboteam/qibo/blob/dbd40109ddbeb9cab4c956be397710e3cb7650a7/doc/source/code-examples/advancedexamples.rst?plain=1#L68-L79

I know it might be a bit of a pain, but citing the Rust book:
<blockquote>

*Nothing is better than documentation with examples. But nothing is worse than examples that don’t work because the code has changed since the documentation was written.*
<p align="right">
<a href="https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#documentation-comments-as-tests">The Rust book, <em>Documentation Comments as Tests</em></a>
</p>
</blockquote>

AleCandido:
Btw, I believe you copied this from Qibo
https://github.com/qiboteam/qibosoq/blob/b2b5d956d21ea5eadcba8ceaeae41095ca9e543e/doc/source/conf.py#L93-L96
but there is no `example/` folder in Qibosoq.

In any case, it seems unharmful, so you could leave it there, and even consider to move some examples there.
https://github.com/qiboteam/qibo/blob/master/doc/source/code-examples/applications.rst?plain=1

It is not necessarily the best layout - I leave this choice to you.
My only points are that tests should run in the CI as much as possible (previous comment) and that we should review even the repo infrastructure periodically (and only keep things that are useful and used).

AleCandido (approved):
Sorry, I thought I already submitted, but it seems like I forgot to press the button...

andrea-pasquale (approved):

PR #97, opened 2023-12-21
Change Black maxline

rodolfocarobene:
For some reason, I was using a max-line-length of 120 in the black pre-commit...
I removed it so now it's back to the default (100)

Checklist before merge:
- [ ] Tests are passing.


AleCandido (approved):
I believe you just copied from Qibolab, I removed it only recently.

Better to just use the default, I agree :)

PR #98, opened 2023-12-22
Update QICK supported version

### LINKED ISSUE #36: Update QICK
now the dependency on qick is fixed to version `0.2.135`, updating it to the newest version (that does not add new capabilities, but does some refactoring) causes some errors

rodolfocarobene:
Closes #36  

I wanted to update the supported version at least for `qick==0.2.165` (since the INFN has the ZCU208 with that qick).

However, downloading qick with
```bash
pip install qick==0.2.165
```
does not fully work... there is a submodule (drivers) that does not get installed even on board.
With `qick===0.2.175` it works just fine until `qick==0.2.181`.

 After it, there are few real breaking changes that I would implement in a new qibosoq version so that until (let's say) 0.1.1 we support < 0.2.181 and then >0.2.181

Anyway, it seems that no real change was made on qick, in fact I could launch a small experiment with no problem. Tests were failing because the local reresentation of the QickSoc objects is changed

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1757,13 +1757,13 @@ cffi = {version = "*", markers = "implementation_name == \"pypy\""}
 
 [[package]]
 name = "qick"
-version = "0.2.175"
+version = "0.2.181"
 description = "Quantum Instrumentation Controller Kit software library"
 optional = false
 python-versions = ">=3.6, <4"
 files = [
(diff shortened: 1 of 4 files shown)


rodolfocarobene:
For the moment I'm leaving the PR as a draft because I want to understand a bit better the problem with 0.2.165 but it should be already ok

rodolfocarobene:
Actually, I would try to merge this and #95 when possible. After that we could do a release I think

AleCandido (approved):

PR #111, opened 2024-03-20
Making RFSoC great again

rodolfocarobene:
Some updates

Checklist before review:
- [ ] Update json for testing
- [ ] Test with all the experiments in the documentation
- [ ] Test with qibocal (patched)

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/src/qibosoq/programs/base.py
+++ b/src/qibosoq/programs/base.py
@@ -56,7 +56,7 @@ class BaseProgram(ABC, QickProgram):
             [elem for elem in self.sequence if elem.type == "readout"]
         )
 
-        self.relax_delay = self.us2cycles(qpcfg.repetition_duration)
+        self.relax_delay = self.us2cycles(qpcfg.relaxation_time)
         self.syncdelay = self.us2cycles(0)
         self.wait_initialize = self.us2cycles(2.0)
 
--- a/doc/source/getting-started/communication.rst
+++ b/doc/source/getting-started/communication.rst
@@ -30,7 +30,7 @@ The dictionary has to contain the following elements:
         "cfg": {
             "soft_avgs": int,
             "reps": int,
-            "repetition_duration": int,
+            "relaxation_time": int,
             "adc_trig_offset": int,
             "average": bool,
         }
--- a/doc/source/getting-started/qibosoq_lab_cal.rst
+++ b/doc/source/getting-started/qibosoq_lab_cal.rst
@@ -254,7 +254,7 @@ For Qibosoq, we need to define the pulses explicitly at the beginning of each ex
 
   sequence = [pulse]
   config = Config(
-            repetition_duration=50, # us
+            relaxation_time=50, # us
             adc_trig_offset=0,
             reps=1,
             soft_avgs=1000,
@@ -389,7 +389,7 @@ For Qibosoq, the experiment needs to be defined from scratch as per the time of 
 
   sequence = [pulse]
   config = Config(
-            repetition_duration=50,
+            relaxation_time=50,
             adc_trig_offset=200,
             reps=1000,
             average=True
@@ -557,7 +557,7 @@ As Qibosoq does not have a way of natively storing results of experiments, the n
   )
 
   config = Config(
-      repetition_duration = 50,
+      relaxation_time = 50,
       reps = 1000
   )
   qubit = Qubit()
@@ -722,7 +722,7 @@ The experiment is similar to the ones before it, we just need to change a couple
   )
 
   config = Config(
-      repetition_duration = 50,
+      relaxation_time = 50,
       reps = 1000
   )
   qubit = Qubit()
@@ -888,7 +888,7 @@ In this case, for example, we can see a new type of sweeper (the delay one) that
   )
 
   config = Config(
-      repetition_duration = 50,
+      relaxation_time = 50,
       reps = 1000
   )
   qubit = Qubit()
@@ -1036,7 +1036,7 @@ For Qibosoq we just need to deactivate the averaging option in the Config object
   sequence = [pulse_1, pulse_2]
 
   config = Config(
-      repetition_duration = 50,
+      relaxation_time = 50,
       reps = 10000,
       average = False
   )
--- a/doc/source/getting-started/tutorials.rst
+++ b/doc/source/getting-started/tutorials.rst
@@ -360,7 +360,7 @@ Now we can define the :class:`qibosoq.components.base.Config` object and our :cl
 .. testcode:: python
 
     config = Config(
-        repetition_duration = 10,
+        relaxation_time = 10,
         reps = 2000
     )
     qubit = Qubit(
--- a/poetry.lock
+++ b/poetry.lock
@@ -38,13 +38,13 @@ files = [
 
 [[package]]
 name = "astroid"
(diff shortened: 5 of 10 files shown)


rodolfocarobene:
Hi @alecandido, could you please have a look at this and #102?
As far as I know, in the lab they have been using #102 for some time without major problems. Eventually just the ones that I fixed in this PR.

I did all the tests I could, something that is missing (i would like to do it today/tomorrow) is testing with qubits, but since the changes are very limited, I think this may be ready for review

src/qibosoq/programs/base.py
```diff
@@ -272,11 +272,28 @@ def collect_shots(self) -> Tuple[list, list]:
...
-            q_val = self.dq_buf[idx].reshape(shape) / lengths[idx]
+            count = int(adc_count[idx])
+            try:
+                shape = (
+                    count,
+                    int(np.prod(self.sweep_axes)),
+                    self.reps,
+                )  # type: Union[Tuple[int, int], Tuple[int, int, int]]
+            except AttributeError:
+                shape = (count, self.reps)
```
alecandido, line 287: Could you add a comment here to explain what is happening (and why you might not have `self.sweep_axes`).

I believe you're inheriting this nice inhomogeneity from `QickProgram` (I didn't check myself). But it would have been nicer if `self.sweep_axes`, for no sweepers, would have just been an empty sequence.

src/qibosoq/programs/base.py
```diff
@@ -272,11 +272,28 @@ def collect_shots(self) -> Tuple[list, list]:
...
+
+            i_val = (
+                self.di_buf[idx][: len_buf // number_of_adcs].reshape(shape)
+                / lengths[idx]
+            )
+            q_val = (
+                self.dq_buf[idx][: len_buf // number_of_adcs].reshape(shape)
+                / lengths[idx]
+            )
```
alecandido, line 296: I know you're delivering `tot_i` and `tot_q`, but it would be nicer to stack them in a single array (with one more dimension, of length 2), such that you won't have to repeat the same operations.

Even if you wanted to keep returning the same values, you could also stack just internally in this function, at least for the time being (it would be a nice compromise for this PR, in such a way that the diff remains limited).

src/qibosoq/programs/base.py
rodolfocarobene: I know that this is a bit weird, I could check if self is NDAveragerProgram, maybe.
Anyway, I added some more specific comment. Something that could be done in the future is unify the shapes for all the different types of acquisition. However this will require some more changes and I would do that in a new release.

src/qibosoq/programs/base.py
rodolfocarobene: For now, I am satisfied of what we have now. I can open a PR to fix all these shape-related problems for results.
I am using different arrays also since I am using di_buf and dq_buf. However I know that there is also a dd_buf that could be helpful, but I have to look into that more

alecandido: Actually, having a look from a broader perspective, you could just stack `self.di_buf` and `self.dq_buf` even outside the loop, that could be kept in NumPy.

The only weird thing is that the shape depends on `idx`...

Is it really possible that you have different ranks? Or are they going to be different just in `counts`?

Otherwise, you could do your processing (divide by `lengths`, cut by `len_buf // number_of_adcs]`) and just reshape at the end.

alecandido: As a quick reply (I could refine the code):
```python
    stacked = np.stack((self.di_buf, self.dq_buf), axis=0)[:, :, :len_acq] / lengths[idx]
    shapes = []
    for idx, adc_ch in enumerate(unique_adcs):
        count = int(adc_count[idx])
        try:
            shapes.append((count, int(np.prod(self.sweep_axes)), self.reps))
        except AttributeError:
            shapes.append((count, self.reps))

    # something like stacked.reshape(shapes)
    return tot_i, tot_q
```

But if you have to really `.reshape()` with different `count`s, the `.reshape()` could be kept in the loop.

rodolfocarobene: Sorry, also lengths[idx] could change so it should be kept inside the loop.
Honestly I am not seeing much more clarity, I think that could came if we use the dd_buf, but otherwise it seems to become just more convoluted without much advantage

alecandido: Whoops, I wrote it too fast.

However, it is fine for `lengths[idx]` to depend on `idx`, since it is just an array.
You can both do:
```py
x = np.ones((20, 10)) * 10
y = np.ones((20, 10)) * np.arange(10)
```
NumPy operations are applied element-wise (here you should adequately compensate for the higher rank with `np.newaxis`, but that's it).

Arrays operations are more expressive, and thus more declarative: you are closer to describe the result you want, rather than the algorithm to obtain.
When applied in Python, they are also faster (just because of Python's slow-down), though I hope here is not a problem.

alecandido: However, what is clearer is definitely subjective. I'm just telling you my experience, and motivating it.
If eventually you like more the current one because you believe there is no better way, that's just the end of the story :)

alecandido: This would be my proposal:
```py
    tot = []
    stacked = np.stack((self.di_buf, self.dq_buf)) / lengths[:, np.newaxis]
    for idx, count in enumerate(adc_count.astype(int)):
        try:
            shape = (count, int(np.prod(self.sweep_axes)), self.reps) 
        except AttributeError:
            shape = (count, self.reps)

		tot.append(stacked[:, idx].reshape(shape))

    return list(zip(*tot))
```

Such that the `for` loop is only restricted to the only thing that you couldn't vectorize (i.e. the reshape, because of the different shapes).

rodolfocarobene: I tried to implement it, but I cannot make mypy happy about the return type (that it sees as tuple[list[any], ...], while ti should be tuple[list, list]). So I added a mypy exception.
I am using lists to send them through the socket (if I remember correctly I cannot do otherwise).

I am still not fully convinced, but I'm trusting you :-)

rodolfocarobene:
@alecandido, I would merge this and the other PR when it is possible. Then wait for some time to give @JavierSerranoGarcia some time for testing on real hardware and then do a release and merge the qibolab PR

src/qibosoq/programs/base.py
alecandido: I tested the output, but in pseudoinput.

If you could provide real input, I would check whether the output is the same before and after.

alecandido (approved):
One way or another, the PR is good to go.

(I forgot to approve before, but you didn't need my approval, since it's not pointing to `main`)

P.S.: I mean, you don't have to merge immediately, and if you prefer you could even roll back and merge as well; I just wanted to let you know that you can merge whenever you wish (though, as usual, I'd merge first #102, unless you have an explicit motivation to avoid it)

rodolfocarobene:
> (I forgot to approve before, but you didn't need my approval, since it's not pointing to `main`)

This I know it, but I wanted to wait for a formal approval to not exploit the system :-)

> As usual, I'd merge first #102, unless you have an explicit motivation to avoid it

I think it's fine, but for #102 I really have to wait for an approval (the changes in that PR, however, are quite minimal and already tested)

PR #113, opened 2024-03-27
Add support for interpolated channel

rodolfocarobene:
I'm adding support for the interpolated channel. 
This is an easy solution (that should work), but I do not like it much... It is not much clean and does not scale considering the addition of new type of channels (what will I do, continuously add new attributes?).
Moreover, it does not solve the problem of having normal readout channels as well as the multiplexed ones.
I'm going to think about it a bit more, but in the meantime this is fine.

Checklist before review:
- [ ] Add support in qibolab
- [ ] Test on hardware

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/src/qibosoq/components/pulses.py
+++ b/src/qibosoq/components/pulses.py
@@ -58,6 +58,9 @@ class Pulse(Element):
     type: str
     """Can be 'readout', 'drive', 'flux'."""
 
+    mixer_frequency: Optional[int] = None
+    """Mixer frequency for interpolated channels."""
+
 
 @dataclass
 class Rectangular(Pulse):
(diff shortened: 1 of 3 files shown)


rodolfocarobene:
Hi @JavierSerranoGarcia, this PR should implement support for interpolated channels.

It has to be tested (with the firmware that you are using at TII) accompanied to this qibolab PR: https://github.com/qiboteam/qibolab/pull/863.

Could you please try and report any error you get? Thanks in advance! :-)

JavierSerranoGarcia:
Thank you @rodolfocarobene , I will test it ASAP

JavierSerranoGarcia:
Hi @rodolfocarobene, I did a quick test and it seemed it works but still not fully tested. Close it and if any problem arises in the  future we will open an issue

PR #115, opened 2024-04-14
Adding RuntimeLoopError in qibosoq.client

rodolfocarobene:
Checklist before review:
- [x] Add tests

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/src/qibosoq/client.py
+++ b/src/qibosoq/client.py
@@ -14,6 +14,14 @@ class QibosoqError(RuntimeError):
     """
 
 
+class RuntimeLoopError(QibosoqError):
+    """Exception raised when qibosoq server encounters a readout loop error.
+
+    Attributes:
+    message -- The error message received from the server (qibosoq)
+    """
+
+
 def connect(server_commands: dict, host: str, port: int) -> Tuple[list, list]:
     """Open a connection with the server and executes the commands."""
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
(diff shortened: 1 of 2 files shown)


src/qibosoq/client.py
```diff
@@ -33,6 +50,11 @@ def connect(server_commands: dict, host: str, port: int) -> Tuple[list, list]:
             received.extend(tmp)
         results = json.loads(received.decode("utf-8"))
         if isinstance(results, str):
+            if "exception in readout loop" in results:
+                raise RuntimeLoopError(results)
+            buffer_overflow = r"buffer length must be \d+ samples or less"
+            if re.search(buffer_overflow, results) is not None:
+                raise BufferLengthError(results)
```
alecandido, line 57: Instead of processing the string, you could add a marker in case of errors.

The easiest could be to define these errors in the server, and register them as an enum. Then you could send something like `{"error": 1, "msg": "the text of my error"}` and deserialize with `ServerErrorEnum[results["error"]](results["msg"])`. Or anything like this.

alecandido (approved):
It is a small addition, and it could be fine like this.

However, I added a suggestion for an alternative implementation.

src/qibosoq/client.py
rodolfocarobene: Wouldn't that just add new complexity layers?
In the server we would still need to check the strings (because the errors are all of the same type). Then we will have the enum, and another part of the client that converts the result back to object of the enum type?

alecandido: Sorry, I didn't read properly the errors, and I just assumed that the check on the buffer length was in Qibosoq, not in QICK.

So, if there isn't any exception generated by Qibosoq itself, yes, it's just overhead (unless QICK is typing its exceptions, in which case we could register those).

rodolfocarobene: Yes, unfortunately it's just QICK :-)

PR #122, opened 2024-06-27
Testing some window functions

rodolfocarobene:
For the moment I'm not sure if it is interesting to merge, but I would try to do some testing to see if control quality present some improvements.

Apparently "Hann functions" are much more used than gaussians for the control (they present less frequency leakage) so it may be interesting also for qibolab (@stavros11, @alecandido) ...

[Wikipedia page about window functions](https://en.wikipedia.org/wiki/Window_function#Hann_and_Hamming_windows)

image

Checklist before merge:
- [ ] Tests are passing.


alecandido:
> Apparently "Hann functions" are much more used than gaussians for the control (they present less frequency leakage) so it may be interesting also for qibolab (@stavros11, @alecandido) ...

@rodolfocarobene on the one side, I'd say: why not?
They could just be dropped in here:
https://github.com/qiboteam/qibolab/blob/0.2/src/qibolab/pulses/envelope.py
and left there for the drivers to support or not (in general, most of them will, since it will be implemented as most of the others: uploading the samples).

On the other side, for the practical use, I believe we're in the typical case of *no free lunches*: to achieve an even smaller leakage over far frequencies, you leak more on the closer ones.
For sure, it's interesting to experiment. But I believe the best optimization should be hardware-aware, there won't be a single one that is working best in all setups. And maybe this will truly enter the domain of optimal control (i.e. just picking a shape trying a few ones, maybe it's not the best option any longer).

But maybe there are some general features that are strong enough to distinguish between these shapes for all chips.

For faster reference:
image
image
image

alecandido:
Btw, pay attention that the Hann in your picture is not centered about the exact same point of Gauss (for Nutall I can't tell). It seems that Hann is slightly shifted to the right (it's a very minor effect, but I didn't expect it).

rodolfocarobene:
I've decided to remove the nuttal function since I didn't find anything specific to that in literature. 
While I wish to merge the Hann that is apparently widely used. The most immediate advantage in respect to the gaussian is the fact that does not require truncation and is still quite simple.

If you agree, I could maybe also add this to qibolab in order to actually add the support for this

alecandido (approved):

alecandido:
> I've decided to remove the nuttal function since I didn't find anything specific to that in literature.

Did you test it in any way?

I had the impression that the Nuttal transformed was a bit weird: despite going down a lot (even more than the Hahn), it had a larger "central body".

rodolfocarobene:
I did test it in rabis, but with our current hardware it was not possible to see any difference... A full RB would be required I think

alecandido:
> I did test it in rabis, but with our current hardware it was not possible to see any difference... A full RB would be required I think

Makes sense

PR #123, opened 2024-07-18
DAC calibration

rodolfocarobene:
This PR adds a notebook (in rst format) that explain how to calibrate DACs in order to get a function that converts a frequency/amplitude/reconstuction-waveform to dBm (@marcogobbo).

I would like to add sooner or later also a notebook with the corresponding ADC calibration, but let's start from here.

Is there a better way of uploading a jupyter notebook, @alecandido? Maybe having this rst is worse than the notebook itself, but I was not sure what the best option was.

Checklist before merge:
- [ ] Tests are passing.


alecandido:
> Is there a better way of uploading a jupyter notebook, @alecandido? Maybe having this rst is worse than the notebook itself, but I was not sure what the best option was.

The rst is not a bad idea, since we're rendering it as docs, and all our Sphinx is made of rst. So, it's one of the best solutions for this

However, the idea is that you should always commit the source of your code, in such a way that you can still modify in the original way (even if it's a different person on a different machine, cloning from scratch).
But if you can convert it back to the original notebook (from files in here), that's also fine. Otherwise, you should commit the notebook as well (and consider generating the rst in Sphinx, with your own script, hooked into `conf.py`, or a suitable Sphinx extension, like [`nbsphinx`](https://nbsphinx.readthedocs.io/en/0.9.3/)).

In general, notebooks are great, but what's the best way to include them is still something that I have to work out. And there could be multiple best solutions, according to the situation.
`nbconvert` (shipped by default with Jupyter) already provide a variety of formats, and there is also [Jupytext](https://jupytext.readthedocs.io/en/latest/index.html).
I encourage you to pick the option that requires the least effort for you, and for someone else that might want to edit the notebook in the future.
(the other relevant bit are diffs, that are uselessly complex with notebooks - but at least GitHub, not Git, is becoming smarter and smarter on that side)

alecandido:
Ok, scrape the rant above: in this case is not going to be included in the docs, so the plain notebook could be the solution.

Otherwise, you could try Jupytext with Python.
But notebooks have the advantage of being rendered by GitHub, and more familiar to people to edit. I'd go for that.

extras/DAC_calibration/Calibration.rst
```diff
@@ -0,0 +1,321 @@
+Calibrate QICK units
+====================
+
+QICK (and therefore Qibosoq) uses arbitrary units to set the amplitudes
+of its pulses. It is not easy to directly connect a specific value of
+amplitude to a physical one in volt or dBm because this depends on two
+factors:
+
+-  The frequency
+-  The reconstruction waveform emplyoed (what the DAC syntesise between
```
alecandido:
```suggestion
-  The reconstruction waveform employed (what the DAC syntesise between
```

extras/DAC_calibration/Calibration.rst
```diff
@@ -0,0 +1,321 @@
...
+   two samples)
+
+In this document we will present a way of calibrating these units from
+the values measured using a Spectrum Analyzer. The final objective is to
```
alecandido: It would be nice to have a Qibolab driver even for that :P

rodolfocarobene:
Makes sense to just use jupyters. At least for "examples" of this sort

extras/DAC_calibration/Calibration.rst
```diff
@@ -0,0 +1,321 @@
+Calibrate QICK units
+====================
+
+QICK (and therefore Qibosoq) uses arbitrary units to set the amplitudes
+of its pulses. It is not easy to directly connect a specific value of
+amplitude to a physical one in volt or dBm because this depends on two
```
alecandido: Not sure about the meaning of expressing a tension in `dBm`. That should be the attenuation, but on top of somehting - and usually that something is the DAC output, so you will need sooner or later a value in Volt.

alecandido (commented):
Ok, let me submit the comments, since the file just changed :P

alecandido:
Unfortunately, GitHub gives nice rendering even for Notebooks diffs, but I can not comment on the lines. So all the review comments will be just on the file.

extras/DAC_calibration.ipynb
alecandido, line 1: Above [3]: RW acronym used, but previously undefined.

(it's in the title, but the first time you mention in the body, make it explicit, and put the acronym aside in parentheses)

extras/DAC_calibration/Calibration.rst
alecandido: Sorry, I confused `dB` and `dBm`. However, why you should measure a pulse in W? Are you assuming the output current to be fixed? Doesn't it also depend on the load?

extras/DAC_calibration.ipynb
alecandido, line 1: [5]: Maybe mention that the order of the polynomial is arbitrary, and explain how did you pick it in this case (i.e. explain why `6`)

alecandido (approved):
I will read it again, and in case provide further suggestions. But in general it seems good and useful.

Thanks @rodolfocarobene!

extras/DAC_calibration/Calibration.rst
rodolfocarobene: dBm are the standard in this type of measurements, I think that's the case beacuse the load is always fixed at 50 Ohm

PR #124, opened 2024-07-19
Python 3.12 and Nix

alecandido:
I was adding Nix files (as for the other repos), but in the process I also noticed that the Python version was stuck before py3.12. Thus, I decided to upgrade even that, and abandon py3.8 (which has been dropped since long in Qibolab).

- [x] add NumPy dependency
  - currently it is hidden: it is used, but undeclared
- [x] fix tests
- [x] fix workflows


alecandido:
@rodolfocarobene sorry, I opened this mainly to add Nix, but it turned out there was a bunch of maintenance overdue :sweat_smile:

rodolfocarobene (approved):

PR #138, opened 2025-04-14
Fix readout length

rodolfocarobene:
FBK noticed this error, we were effectively using the wrong number of cycles for the acquisiton since we were using the wrong clock.
This was only noticable with raw_acquisition, but was always present


alecandido (approved):
Great the error has been spotted and fixed!

However, it may be even better to expand a bit the docstring, or introduce a brief comment. These are all functions and elements that are mostly exposed to QICK. But even QICK is not NumPy, so it could be optimistic to overly rely on their docs. Being slightly redundant on our side should not be a huge price.

PR #139, opened 2025-04-15
Add support for external clock reference

rodolfocarobene:
This is also an addition suggested by FBK that is currently using a patched version of qibosoq for this feature.
Anyway this should work, but I'd like to briefly test it before merging

Checklist before merge:
- [ ] Tests are passing.

### DIFF
--- a/doc/source/getting-started/usage.rst
+++ b/doc/source/getting-started/usage.rst
@@ -23,7 +23,8 @@ A list of the configurable variables (with values set to the default ones) is th
    export QIBOSOQ_BITSTREAM=/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/qick_111.bit
    # is the readout multiplexed?
    export QIBOSOQ_IS_MULTIPLEXED=False
-
+   # is an external clock used as a reference?
+   export QIBOSOQ_EXT_CLK=False
 
 Running the server
 """"""""""""""""""
@@ -80,3 +81,14 @@ Some examples are:
           echo "No running server"
       fi
     }
+
+External clock reference
+""""""""""""""""""""""""
+
+It is possible to provide an external clock reference to the board, allowing all clocks to be synchronized with other instruments.
+
+- For the **RFSoC4x2**, connect a 10 MHz reference to the **CLK_IN** SMA input.
+- For the **ZCU111**, provide a 12.8 MHz reference on **External_REF_CLK**.
+- For the **ZCU216**, supply a 10 MHz reference to **INPUT_REF_CLK**.
+
+Note that, for the external reference to be detected, the environment variable `QIBOSOQ_EXT_CLK` must be set to `True`.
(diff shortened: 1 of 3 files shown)


src/qibosoq/client.py
```diff
@@ -71,6 +73,17 @@ def convert_commands(obj_dictionary: dict) -> dict:
         dict_dictionary["sweepers"] = [
             sweep.serialized for sweep in obj_dictionary["sweepers"]
         ]
+
+        # In sweepers all the pulses are registered in initialitialization (before loop)
+        # this means that if the same DAC has multiple values (pulses) they get overwritten
```
alecandido, line 78: I would say that restricting to one pulse per-DAC in a swept sequence is a further constraint.
Or, at least, I am missing myself the connection to the registration at initialization time...

alecandido (commented):
The external clock part is mostly fine. Just a comment on boolean env vars.

Instead, I'm struggling to understand the further checks on sweepers (which we could even split to another PR, since they seem unrelated to the external clock part).

src/qibosoq/client.py
rodolfocarobene: Yes, exactly. The behavior until now was that the sequence was executed but was different from the one defined. Because of this I added an explicit error.

doc/source/getting-started/usage.rst
```diff
@@ -23,7 +23,8 @@ A list of the configurable variables (with values set to the default ones) is th
    export QIBOSOQ_BITSTREAM=/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/qick_111.bit
    # is the readout multiplexed?
    export QIBOSOQ_IS_MULTIPLEXED=False
-
+   # is an external clock used as a reference?
+   export QIBOSOQ_EXT_CLK=False
```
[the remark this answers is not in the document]
rodolfocarobene: This is correct, maybe I should use something like "value in (True, true, 1)" instead of "value == True"

src/qibosoq/client.py
rodolfocarobene: It is Qibosoq related, to overcome it I should assign one or more new registers for the parameter swept.

I do something similar for the delay and in principle it is extendable.
It just never was a priority though, since for calibration experiments this rarely happens.

Example of what I should do:
```python
reg_1 = 1
reg_2 = 2

def loop():
	amp = reg_1
	pulse(amplitude=amp)
	amp = reg_2
	pulse(amplitude=amp)

def update_each_loop():
	reg_2 += 1
```

alecandido: > It is Qibosoq related, to overcome it I should assign one or more new registers for the parameter swept.

Actually, I'm doing something like that for Q1ASM (Qblox's assembly) generation. Now I wonder to which extent we could even reuse it :thinking:

doc/source/getting-started/usage.rst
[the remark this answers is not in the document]
alecandido: Perfect like this :)

PR #141, opened 2025-06-04
Bump versions of `ipython`, `jinja2` and `tornado`

renatomello:
`jinja` and `tornado` are updated via `nbsphinx`


alecandido (approved):

alecandido:
Tests are failing for a seemingly unrelated reason

rodolfocarobene:
we wanted to support python 3.9 because some of the boards may still have an old version of pynq. 
However I believe now it is fine to abandon it

alecandido:
> @alecandido @rodolfocarobene I was wondering if there was any reason not to support `python 3.13`, so I enabled it and the tests are all passing. Is there any external reason?

Not that I'm aware of.

In Qibolab it was just blocked by QUA (and it is not any longer), but this is completely independent

PR #142, opened 2025-07-09
Restore support for Python 3.8

alecandido:
Apparently, QICK only ships images with py3.8.
Despite being possible to update it, it seems to be enough inconvenient for users.

In any case, we should have almost no issue in keep support for py3.8. At most, it will require to revert usage of some new syntax.

Most of the problem is faced with old dependencies, which most likely have also been flagged with some securities' vulnerability (I may remember something related to IPython). But there is almost no security concern for Qibosoq in general, and most of the dependencies are actually build dependencies. While at runtime Qibosoq just relies on standard library, QICK, and NumPy.

However, notice that, exactly because of the NumPy, Qibosoq will be prevented to be used with recent versions of Qibo and Qibolab (to support Python 3.8, you need at most NumPy v1.24, and not beyond - while Qibo currently dropped Numpy v1, and it is formally only compatible with NumPy v2).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,30 +2,32 @@
   "nodes": {
     "cachix": {
       "inputs": {
-        "devenv": "devenv_2",
+        "devenv": [
+          "devenv"
+        ],
         "flake-compat": [
+          "devenv"
+        ],
+        "git-hooks": [
           "devenv",
-          "flake-compat"
+          "git-hooks"
         ],
         "nixpkgs": [
           "devenv",
           "nixpkgs"
-        ],
-        "pre-commit-hooks": [
-          "devenv",
-          "pre-commit-hooks"
         ]
       },
       "locked": {
-        "lastModified": 1712055811,
-        "narHash": "sha256-7FcfMm5A/f02yyzuavJe06zLa9hcMHsagE28ADcmQvk=",
+        "lastModified": 1748883665,
+        "narHash": "sha256-R0W7uAg+BLoHjMRMQ8+oiSbTq8nkGz5RDpQ+ZfxxP3A=",
         "owner": "cachix",
         "repo": "cachix",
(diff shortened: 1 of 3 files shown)


alecandido:
Yes, but consider that it is quite possible to support both Numpy v1 and v2. Most likely, there is no problem in doing that with the present Qibosoq.

So, the proposal would be:
- fix the current workflows
- extend again the workflows down to py3.8 (hoping that GitHub will keep support for it for another while)
- extend the version range for NumPy as well (allowing v2, but also v1.24, or even previous versions)

The lock file should seal NumPy v1.24 (or former). But this has no impact on the user, which will be able to install even NumPy v2, if not ruled out.
We may have had a problem with locking, since there is no NumPy version covering the entire Python range. But this already worked:
https://github.com/qiboteam/qibosoq/blob/8d791e62054c5c3bbbe29c37fcb06ede61bf4f04/poetry.lock#L1639-L1646
https://github.com/qiboteam/qibosoq/blob/8d791e62054c5c3bbbe29c37fcb06ede61bf4f04/poetry.lock#L1678-L1685
since Poetry solved it on its own, splitting the versions with this "markers" (we could even try to add markers explicitly, and impose NumPy v2.0 as lower bound if the Python version is suitable - but I'd say that there is truly no purpose in doing that, since Qibosoq does not really need NumPy v2 at all - let's leave it to Qibo and Qibolab).

alecandido:
> If your solution works, It's fine for me

Unfortunately, I'm a bit busy right now (let's say, until next Thursday). As soon as I will have some time, I will try to fix the workflows, as described above.

rodolfocarobene:
I see, I'll try to give it a go myself (in a new PR point at this) to help the process :-)

rodolfocarobene (approved):

PR #143, opened 2025-07-11
Add support >= python 3.8 (second)

rodolfocarobene:


pyproject.toml
```diff
@@ -36,6 +37,8 @@ optional = true
 [tool.poetry.group.analysis.dependencies]
 pylint = ">=2.16.0"
 mypy = "^1.10.1"
+tomli = { version = "^2.2.1", python = "<3.11" }
+dill = { version = ">=0.3.6", python = "<3.11" }
```
alecandido, line 41: Do we actually need `dill`?

It is extending `pickle`, which is standard library. And while it could be used as a drop-in replacement of `pickle`, `dill` itself has not become part of standard library.

Where is this used?

The only reference to `pickle` I can find in the 3.11 changelog is

> Added [object.__getstate__()](https://docs.python.org/3.11/library/pickle.html#object.__getstate__), which provides the default implementation of the __getstate__() method. [copy](https://docs.python.org/3.11/library/copy.html#module-copy)ing and [pickle](https://docs.python.org/3.11/library/pickle.html#module-pickle)ing instances of subclasses of builtin types [bytearray](https://docs.python.org/3.11/library/stdtypes.html#bytearray), [set](https://docs.python.org/3.11/library/stdtypes.html#set), [frozenset](https://docs.python.org/3.11/library/stdtypes.html#frozenset), [collections.OrderedDict](https://docs.python.org/3.11/library/collections.html#collections.OrderedDict), [collections.deque](https://docs.python.org/3.11/library/collections.html#collections.deque), [weakref.WeakSet](https://docs.python.org/3.11/library/weakref.html#weakref.WeakSet), and [datetime.tzinfo](https://docs.python.org/3.11/library/datetime.html#datetime.tzinfo) now copies and pickles instance attributes implemented as [slots](https://docs.python.org/3.11/glossary.html#term-__slots__). This change has an unintended side effect: It trips up a small minority of existing Python projects not expecting [object.__getstate__()](https://docs.python.org/3.11/library/pickle.html#object.__getstate__) to exist. See the later comments on [gh-70766](https://github.com/python/cpython/issues/70766) for discussions of what workarounds such code may need. (Contributed by Serhiy Storchaka in [bpo-26579](https://bugs.python.org/issue?@action=redirect&bpo=26579).)
>
> https://docs.python.org/3.11/whatsnew/3.11.html#other-language-changes

alecandido (approved):
Other than understanding where `dill` is coming from, all changes are fine with me.

Thanks! @rodolfocarobene

PR #144, opened 2025-07-16
clean grouping of mux readouts

### LINKED ISSUE #128: Multiplexed readout grouping
https://github.com/qiboteam/qibosoq/blob/f51693873575251eea5bbfab9998f4977c17aced/src/qibosoq/programs/base.py#L346-L363

The goal of the former function would be to group together simultaneous readout pulses.

The concept of simultaneity for pulses may be ambiguous, since any kind of overlap is possible. In this case, the definition suggested by the function would be that they are simultaneous if *any* kind of overlap is happening (notice that this is not a transitive relation, so you could end up grouping together two non-simultaneous pulses, just because they are both simultaneous to a third one *in the middle*).

However, the current code is not even implementing this solution, since it is comparing the length of the previous readout pulse with the start delay with respect to the previous pulse (i.e. any kind of pulse). This means that a `readout - drive - readout` may end up making the two `readout` pulses simultaneous, when they are not even overlapping to the same one, e.g.:
```
r1: ############
d1:                     ####
r2:                          ############
```
(the difference `r2.start - d1.start` is definitely shorter than `r1.length`)

In general, the simultaneity determined by Qibosoq is not reliable.

I'm not sure what are the constraints imposed by the QICK firmware, but on the one side, whatever it is doing should be reproduced. If there is no QICK-related constraint, I would give up on simultaneity, and just have a flat list of readouts.

rodolfocarobene:
Closes #128 

My solution in the end is to first group the readout if they start exactly at the same time and then check if they share the same duration.
I am letting QICK handle other type of exceptions

### DIFF
--- a/src/qibosoq/programs/base.py
+++ b/src/qibosoq/programs/base.py
@@ -359,17 +359,14 @@ class BaseProgram(ABC, QickProgram):
         group: List[Element] = []
 
         for pulse in self.sequence:
-            if pulse.type != "readout":
+            readout = pulse.type != "readout"
+            if readout or pulse.start_delay != 0:
                 if group:
                     mux_list.append(group)
                     group = []
-                continue
 
-            if pulse.start_delay != 0:
-                if group:
-                    mux_list.append(group)
-                    group = []
-            group.append(pulse)
+            if !readout:
+                group.append(pulse)
 
         if group:
             mux_list.append(group)
--- a/src/qibosoq/programs/base.py
+++ b/src/qibosoq/programs/base.py
@@ -357,7 +357,7 @@ class BaseProgram(ABC, QickProgram):
         """
         mux_list: List[List[Element]] = []
 
-        group = []
+        group: List[Element] = []
         for pulse in self.sequence:
             if pulse.type == "readout":
                 if pulse.start_delay != 0:
--- a/src/qibosoq/programs/base.py
+++ b/src/qibosoq/programs/base.py
@@ -356,28 +356,31 @@ class BaseProgram(ABC, QickProgram):
         Readout pulses are considered to be correctly organized.
         """
         mux_list: List[List[Element]] = []
-
         group: List[Element] = []
+
         for pulse in self.sequence:
-            if pulse.type == "readout":
-                if pulse.start_delay != 0:
-                    if len(group) != 0:
-                        mux_list.append(group)
-                        group = []
-                    group = [pulse]
-                else:
-                    group.append(pulse)
-        if len(group) != 0:
+            if pulse.type != "readout":
+                if group:
+                    mux_list.append(group)
+                    group = []
+                continue
+
+            if pulse.start_delay != 0:
+                if group:
+                    mux_list.append(group)
+                    group = []
+            group.append(pulse)
+
+        if group:
             mux_list.append(group)
 
         # check that all the readout pulses share the same duration
         for group in mux_list:
-            length = group[0].duration
-            for pulse in group[1:]:
-                if not length == pulse.duration:
-                    raise RuntimeError(
-                        f"Muxed readout pulses must share same duration.: {mux_list}"
-                    )
+            durations = {pulse.duration for pulse in group}
+            if len(durations) > 1:
+                raise RuntimeError(
+                    f"Muxed readout pulses must share same duration.: {mux_list}"
+                )
         return mux_list
 
     @abstractmethod
--- a/src/qibosoq/programs/base.py
+++ b/src/qibosoq/programs/base.py
@@ -359,13 +359,13 @@ class BaseProgram(ABC, QickProgram):
         group: List[Element] = []
 
         for pulse in self.sequence:
-            readout = pulse.type != "readout"
-            if readout or pulse.start_delay != 0:
+            readout = pulse.type == "readout"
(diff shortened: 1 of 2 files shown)


src/qibosoq/programs/base.py
```diff
@@ -352,18 +352,32 @@ def group_mux_ro(self) -> list:
...
+        group: List[Element] = []
+        for pulse in self.sequence:
+            if pulse.type == "readout":
+                if pulse.start_delay != 0:
+                    if len(group) != 0:
+                        mux_list.append(group)
+                        group = []
+                    group = [pulse]
+                else:
+                    group.append(pulse)
```
alecandido, line 369: Uhm, I'm not yet convinced: I believe here you're to discriminate if the new pulse start simultaneously with the current group. But is `pulse.start_delay`:
- relative to the previous readout pulse in the sequence? (which should be what you need)
- or relative to the last pulse in the same channel?
- or relative to the beginning of the experiment itself?

alecandido (commented):
What you're trying to support are experiments in which you read multiple qubits simultaneously, I imagine.

I don't know why you need to manually group them, since I'm not familiar with QICK handling (essentially, the expected input and workflow of the tproc).
However, assuming you somehow need to group the pulses to be played simultaneously, I still see a lot of problems.

Even if done right, you now expect all qubits readout to start at the same time. However, if you're performing a simultaneous Rabi amplitude experiment (just an example), the readout has to come right after a drive pulse. Which may have different duration for the different qubits. You can also consider other experiments involving calibrated drive pulses before a readout.

Moreover, even if we could assume all of them starting at the same time, having all the same duration could be a problem on its own. Whenever you start optimizing the readout pulses, you may also use different readout durations per-qubit.

And everything explodes the moment you would come to mid-circuit measurements (not yet there even in Qibolab).

So, I may have made wrong assumptions above, and thus got to an incorrect conclusion. But rather than solving the grouping algorithm, I'd rather focus on the need for grouping, to better establish (and possibly document) why we need it.

src/qibosoq/programs/base.py
rodolfocarobene: It is relative to the last pulse in the sequence (readout or drive). 
So if the start_delay is not zero we need to create a new readout group for sure, otherwise we just add the measurement to the group (that can be empty or not depending on the type of the last pulse

rodolfocarobene: You are right, I will implement it asap!

alecandido:
The last review is just on the code, but before merging this, we should face the "big" discussion above (about the motivations behind this feature).

src/qibosoq/programs/base.py
```diff
@@ -352,18 +352,32 @@ def group_mux_ro(self) -> list:
...
+
+        for pulse in self.sequence:
+            readout = pulse.type != "readout"
+            if readout or pulse.start_delay != 0:
+                if len(group) > 0:
+                    mux_list.append(group)
+                    group = []
+
+            if not readout:
+                group.append(pulse)
```
alecandido, line 369: Reading this again, the variable name sounds a bit misleading. I'd propose to negate it
```suggestion
            readout = pulse.type == "readout"
            if (not readout) or pulse.start_delay != 0:
                if len(group) > 0:
                    mux_list.append(group)
                    group = []

            if readout:
                group.append(pulse)
```

src/qibosoq/programs/base.py
```diff
@@ -352,18 +352,32 @@ def group_mux_ro(self) -> list:
...
+        if len(group) > 0:
+            mux_list.append(group)
+
+        # check that all the readout pulses share the same duration
+        for group in mux_list:
+            durations = {pulse.duration for pulse in group}
+            if len(durations) > 1:
+                raise RuntimeError(
+                    f"Muxed readout pulses must share same duration.: {mux_list}"
+                )
```
alecandido, line 380: Now that I know the limitation, there is another question arising. The present check does not exclude overlapping readouts: are they allowed or not?

I.e.
```
0/readout  ########
1/readout  ########
2/readout      #########
```
is this valid?

And: what happens if an invalid sequence is fed to QICK?

According to the combination of answers, I would do one of the following:
- **overlapping groups are valid:** then, it could be worth putting in different groups even pulses with different durations, considering them as overlapping, but not part of the same group (possibly giving up some optimization - which would be kept if the duration is the same)
- **QICK validates sequences:** then, we can give up validation on our side, as redundant, and let QICK's error bubble up
- **neither of the above:** then, we should validate and raise errors even for overlapping groups

The last point it's the most cumbersome, but it may be still simple to check during groups creation, checking if the present group duration exceeds the `.start_delay`.
This will require that, instead of closing readout groups even on non-readout pulses, we should just disregard them. However, to avoid incurring in the former problem of `.start_delay` reset, we should maintain a *readout start delay* (even named just `delay`, which is clear given the context), and just increment it on non-readout pulses. So that the loop above will look like

```py
delay = 0
for pulse in self.sequence:
    delay += pulse.start_delay
    if pulse.type != "readout":
        continue

    if delay != 0 and len(group) > 0:
        if delay < group[0].duration:
            raise ValueError()
        mux_list.append(group)
        group = []
        delay = 0

    group.append(pulse)
```

(possibly still rearranged a bit, since otherwise it will at least get repeated for the final group)

Issue #145, opened 2025-07-18
Installation issue with python 3.8 on latest main build


cpcepa:
Thanks for backporting support for Python 3.8. However, I'm experiencing this issue when installing on Python 3.8 (on both my PC and ZCU216), but Python 3.10 installs normally. Using Main branch [4ac7bac].

Any advice?

```
$ pip install .
Processing /ssd_data/chuping/qibosoq
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  ? exit code: 1
  ??> [14 lines of output]
      Traceback (most recent call last):
        File "/home/chuping/miniforge3/envs/qibo_38/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
          main()
        File "/home/chuping/miniforge3/envs/qibo_38/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "/home/chuping/miniforge3/envs/qibo_38/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 164, in prepare_metadata_for_build_wheel
          return hook(metadata_directory, config_settings)
        File "/tmp/pip-build-env-d8u4yqrr/overlay/lib/python3.8/site-packages/poetry/core/masonry/api.py", line 42, in prepare_metadata_for_build_wheel
          poetry = Factory().create_poetry(Path(".").resolve(), with_groups=False)
        File "/tmp/pip-build-env-d8u4yqrr/overlay/lib/python3.8/site-packages/poetry/core/factory.py", line 60, in create_poetry
          raise RuntimeError("The Poetry configuration is invalid:\n" + message)
      RuntimeError: The Poetry configuration is invalid:
        - The fields ['authors', 'description', 'name', 'version'] are required in package mode.
      
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
??> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details. 
```

cpcepa:
[...]

PR #154, opened 2025-09-12
Developer mode installation script

alecandido:
After discussion with @rodolfocarobene, this partly reconsiders the proposed solution for #145.

In particular, in https://github.com/qiboteam/qibosoq/issues/145#issuecomment-3087827498 I proposed to build the package with one Python version - recent enough to be compatible with Poetry 2, required for the current manifest layout - and then install with a different one - old enough to be supported by PYNQ.

While this is a working solution, it is terribly annoying when working in development mode on the board, since it requires working with two different python versions (the recent one to be installed separately), and reinstall every time a change wants to be tested. Or do the same while developing on a separate machine, always uploading the built package there.

Thus, this PR provides a script which bypass Poetry for installation, extracting the dependencies from the manifest, and installing them with bare `pip`.
At the same time, it generates a `qibosoq.pth` file in the environment, the same which would have been provided by `pip install -e` or `poetry install`. This file redirects Python imports, effectively adding the project's folder to the searched library path.

### DIFF
--- a/dev-install.bash
+++ b/dev-install.bash
@@ -1,0 +1,11 @@
+prog=$(
+	cat <<'EOF'
+		/^\]/ { if (s) e=1 }
+		s && !e { match($0, /".*"/); print substr($0, RSTART+1, RLENGTH-2) }
+		/^dependencies/ { s=1 }
+EOF
+)
+awk "$prog" pyproject.toml >qibosoq-requirements.txt
+
+pip install -r qibosoq-requirements.txt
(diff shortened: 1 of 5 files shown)


rodolfocarobene (approved):
It now works!
I had to add the try-except block in __init__ because by only adding the pth file it doesn't load the distribution info and that line fails.
Also, to uninstall it is necessary to manually remove the file so I added a script for that too.

